PR #85, opened 2013-08-27
Make IdentityCache return only readonly records

airhorns:
@fbogsany @camilo @arthurnn for review please.

IdentityCache is a forever inconsistent datastore. It is a duplication of state across two different storage backends, and unless we somehow confirmed commit in both places before acknowledging, we can't guarantee that the two are in sync. The ruby process could die before it has a chance to issue the memcache DEL, the request could get lost if the connection is interrupted or any number of other things could prevent IDC from reflecting whats truly in the database.

This being the case, it is sketchy to write something you find in IDC back to the database, because you might be writing out of date data to the database. This is already the case with a MySQL insert in Rails cause you had to have read that data a finite amount of time ago, but that window is usually much smaller, and we have locking primitives and transactions to manage this. With IDC, we don't have that, and we don't have any way of knowing if the data is out of date without asking the database, which entirely defeats the purpose of IDC.

So, I suggest we return only readonly records from IDC. If you are gonna write a record, first fetch its values from the DB.

### DIFF
--- a/lib/identity_cache/belongs_to_caching.rb
+++ b/lib/identity_cache/belongs_to_caching.rb
@@ -29,7 +29,7 @@ module IdentityCache
             if IdentityCache.should_cache? && #{options[:foreign_key]}.present? && !association(:#{association}).loaded?
               self.#{association} = #{options[:association_class]}.fetch_by_id(#{options[:foreign_key]})
             else
-              #{association}
+              load_and_readonlyify(:#{association})
             end
           end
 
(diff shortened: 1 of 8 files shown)


airhorns:
@camilo do you care about this at all?

rafaelfranca:
@dylanahsmith I think after https://github.com/Shopify/identity_cache/pull/234 there is no reason to not merge this. Can I go ahead and do it?

dylanahsmith:
Last I checked this would cause a lot of test failures in Shopify.  You could test it on a branch with CI if you want to see if that has changed

PR #137, opened 2014-03-17
Adds cache invalidation on class.update_counters

jnormore:
Invalidate cache on update_counters class method. This method takes an array of ids so we fetch and expire cache for each.

@dylanahsmith @tjoyal can you take a look please, thanks!

cc @ViSovari @disaacs

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -72,7 +72,24 @@ module IdentityCache
         records
       end
 
+      def force_expiration(id)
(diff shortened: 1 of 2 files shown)


rafaelfranca:
@jnormore can we delete this branch?

camilo:
I deleted it but meh branches are cheap

rafaelfranca:
I asked more because I was looking if there are any work worth to continue.

camilo:
Oh I see we can restore if that is the case, but since it is abandoned I think this might not be needed anymore

PR #171, opened 2014-08-01
Document cache_belongs_to.

BMorearty:
Added documention in the README for the `cache_belongs_to` statement.

I don't understand some of the options so some of them aren't well-documented. But at least the statement itself is now documented, making it clear that it's possible to cache a `belongs_to` association.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,20 @@ _[:inverse_name]_ Specifies the name of parent object used by the association. T
 Example:
 `cache_has_one :configuration, :embed => true`
 
+#### cache_belongs_to
+
+Options:  
+_[:cached_accessor_name]_ Specifies the name of the method to fetch the associated object. Defaults to `fetch_#{association}`.
+
+_[:foreign_key]_ Specifies the foreign key to use.
+
+_[:association_class]_ Specifies the association class.
+
+_[:prepopulate_method_name]_ Default: `prepopulate_fetched_#{association}`.
+
+Example:
+`cache_belongs_to :database_server`
+
 #### cache_attribute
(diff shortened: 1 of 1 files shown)


TrangPham:
I agree this should be adde to the documentation

README.md
```diff
@@ -183,6 +183,20 @@ _[:inverse_name]_ Specifies the name of parent object used by the association. T
 Example:
 `cache_has_one :configuration, :embed => true`
 
+#### cache_belongs_to
+
+Options:
```
dylanahsmith: Wow, I didn't realized how much we let users override from the options until now.  Most of these options can also be overriden on other configuration dsl methods, but are undocumented, so I'm not sure if they were all intentionally exposed.  I don't think we should allow them to be overridden at all in the future, so I don't think we should start documenting them now.

dylanahsmith:
It looks like we are also missing documentation in the code for cache_belongs_to and have no description of it in the README.

TrangPham:
@dylanahsmith I'm not sure if there was a reason that the cache_belongs_to is in its own file instead of in ConfirgurationDsl. Is there a design reason for this? Or should it be moved into there and cleaned up?

dylanahsmith:
I think it was put in its own file because it doesn't share a lot of the functionality as cache_has_one and cache_has_many.  It is an internal separation of concerns, just like the [other modules that get included along with IdentityCache](https://github.com/Shopify/identity_cache/blob/6916eff952036657316e8497eb5b7aaeffc1dede/lib/identity_cache.rb#L43-47)

PR #179, opened 2014-09-15
should not log when id is not primary key for the model

trumant:


lib/identity_cache/query_api.rb
```diff
@@ -28,7 +28,9 @@ def fetch_by_id(id)
             object = nil
             coder = IdentityCache.fetch(rails_cache_key(id)){ coder_from_record(object = resolve_cache_miss(id)) }
             object ||= record_from_coder(coder)
-            IdentityCache.logger.error "[IDC id mismatch] fetch_by_id_requested=#{id} fetch_by_id_got=#{object.id} for #{object.inspect[(0..100)]} " if object && object.id != id.to_i
+            if object && primary_key == "id" && object.id != id.to_i
```
dylanahsmith: `object.id` returns the primary key value and this check seems appropriate no matter that the primary key is.  I think the problem you are running into is the `id.to_i` if you have a non-integer primary key.

PR #193, opened 2014-12-22
Allow cache invalidation in transactions again.

dylanahsmith:
cc @jstorimer 

## Problem

IDC v0.2.2 has a regression where touching a record no longer invalidates the cache for that record.  This was because the record was being invalidated in an after_touch callback, where `IdentityCache.should_update_cache?` returned falsey, since `IdentityCache.should_update_cache?` returns falsey.

## Solution

Ideally we would always invalidate the cache in an after_commit callback, but invalidating the cache should always be a safe thing to do, even inside a transaction.

The actual concern in updating the cache with `IdentityCache.readonly = true` in our use case was database slave lag, where is it safe to fetch from the cache, but unsafe to fill the cache, since it could fill the cache with stale data.  Therefore, I renamed `should_update_cache?` to `should_fill_cache?` and had it stop requiring `should_use_cache?` to return truthy.

### DIFF
--- a/lib/identity_cache.rb
+++ b/lib/identity_cache.rb
@@ -69,8 +69,8 @@ module IdentityCache
       @logger || Rails.logger
     end
 
-    def should_update_cache? # :nodoc:
-      !readonly && should_use_cache?
+    def should_fill_cache? # :nodoc:
+      !readonly
     end
 
     def should_use_cache? # :nodoc:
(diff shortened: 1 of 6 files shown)


dylanahsmith:
@arthurnn I agree with @fbogsany that `MemoizedCacheProxy` isn't really the right place to implement this.  If we only have a read-only connection to the database, then we can still memoize the values in memory even if we don't want to persist them to memcache.

arthurnn:
that make sense.. lets merge this!

PR #194, opened 2014-12-24
Fix a test that expected expiry without changes to saved record.

dylanahsmith:
@arthurnn for review

In rails 4.2 the parent record isn't touched through the `touch: true` option on the belongs_to association if there were no changes to the record being saved, since no update statement is sent to the database.


BlakeMesdag:
@dylanahsmith is this still good to go?

PR #199, opened 2015-01-17
Refactor and deduplicate parent cache expiry.

dylanahsmith:
@BlakeMesdag & @tjoyal for review

## Problem

We noticed redundant memcache delete requests after pull #184 was introduced.  It looks like this happens when a record embeds a cache blob more than once.

## Solution

When a child expires its parent caches, it first builds up a complete set of all the records that need their cache invalidated, then expires the cache on those records.  A hash is used to implement this set, and the primary cache key is used as a key, so it will avoid adding multiple records to the set for the same database row, even if there are multiple record instances for that row in memory.


lib/identity_cache/configuration_dsl.rb
```diff
@@ -270,23 +270,19 @@ def attribute_dynamic_fetcher(attribute, fields, values) #:nodoc:
...
-          after_commit :#{after_action_name}
-          after_touch  :#{after_action_name}
-          add_parent_expiration_entry :#{after_action_name}
+        child_class.parent_expiration_entries[options[:inverse_name]] << [self, options[:only_on_foreign_key_change]]
 
-          def #{after_action_name}
-            expire_parent_cache_on_changes(:#{options[:inverse_name]}, '#{foreign_key}', #{self.name}, #{options[:only_on_foreign_key_change]})
+        child_class.class_eval(<<-CODE, __FILE__, __LINE__ + 1)
+          after_commit :expire_parent_caches
+          if Gem::Version.new(ActiveRecord::VERSION::STRING) < Gem::Version.new("4.0.4")
```
BlakeMesdag: Might have to also add `ActiveRecord::VERSION::MAJOR` < 4 for rails 3.2 compatibility.

dylanahsmith: This is the same check as we have in [query_api.rb](https://github.com/Shopify/identity_cache/blob/1abbfcb007bc2d5a193e93a8cf7bbbdf9b219a71/lib/identity_cache/query_api.rb#L7).  Also, this will be used for rails 3.2, since it is less then version 4.0.4.

lib/identity_cache/configuration_dsl.rb
```diff
@@ -270,23 +270,19 @@ def attribute_dynamic_fetcher(attribute, fields, values) #:nodoc:
...
         child_class = options[:association_class]
-        child_association = child_class.reflect_on_association(options[:inverse_name])
-        raise InverseAssociationError unless child_association
-        foreign_key = child_association.association_foreign_key
+        raise InverseAssociationError unless child_class.reflect_on_association(options[:inverse_name])
 
         child_class.send(:include, ArTransactionChanges) unless child_class.include?(ArTransactionChanges)
         child_class.send(:include, ParentModelExpiration) unless child_class.include?(ParentModelExpiration)
 
         after_action_name = "expire_parent_cache_#{self.name.underscore}"
```
tjoyal: No longer used

lib/identity_cache/configuration_dsl.rb
```diff
@@ -270,23 +270,19 @@ def attribute_dynamic_fetcher(attribute, fields, values) #:nodoc:
...
-        child_class.class_eval(<<-CODE, __FILE__, __LINE__ + 1)
-        
-          after_commit :#{after_action_name}
-          after_touch  :#{after_action_name}
-          add_parent_expiration_entry :#{after_action_name}
+        child_class.parent_expiration_entries[options[:inverse_name]] << [self, options[:only_on_foreign_key_change]]
 
-          def #{after_action_name}
-            expire_parent_cache_on_changes(:#{options[:inverse_name]}, '#{foreign_key}', #{self.name}, #{options[:only_on_foreign_key_change]})
+        child_class.class_eval(<<-CODE, __FILE__, __LINE__ + 1)
```
tjoyal: Should this now be in `ParentModelExpiration` now that there is no longer any custom code generation?

dylanahsmith: Ya, I guess we can do the `after_commit` when it is included.

tjoyal:
2 tiny comments. I think everything is good, but I'll have to admin these inner working of AR aren't my strong element. It's a :+1: for me but __if__ any of you have an uneasy feeling, I suggest asking for a 3th reviewer.

dylanahsmith:
@arthurnn want to review this as well?

lib/identity_cache/configuration_dsl.rb
arthurnn: should we care at all about this now?
I just dropped tests for 3.2 on master, I think we dont need to support 4.0.4 too.. suporting latest 4.0.x is good enough

lib/identity_cache/parent_model_expiration.rb
```diff
@@ -4,44 +4,61 @@ module ParentModelExpiration # :nodoc:
...
+          end
+          old_parent = klass.find(transaction_changed_attributes[foreign_key])
         rescue ActiveRecord::RecordNotFound => e
           # suppress errors finding the old parent if its been destroyed since it will have expired itself in that case
         end
       end
 
-      true
+      cached_associations.each do |parent_class, only_on_foreign_key_change|
+        if new_parent.try(:is_a?, parent_class) && should_expire_identity_cache_parent?(foreign_key, only_on_foreign_key_change)
```
arthurnn: can we use `new_parent && new_parent.is_a? parent_class` instead of try? (not a big deal tho)

lib/identity_cache/parent_model_expiration.rb
```diff
@@ -4,44 +4,61 @@ module ParentModelExpiration # :nodoc:
...
+      parents_to_expire = {}
+      add_parents_to_cache_expiry_set(parents_to_expire)
+      parents_to_expire.each_value do |parent|
+        parent.send(:expire_primary_index)
       end
     end
 
-    def expire_parent_cache_on_changes(parent_name, foreign_key, parent_class, only_on_foreign_key_change)
-      new_parent = send(parent_name)
+    def add_parents_to_cache_expiry_set(parents_to_expire)
```
arthurnn: `parent_expiration_entries` is not a Set.. not sure if we should have the suffix `_set` on the method name

lib/identity_cache/configuration_dsl.rb
dylanahsmith: Looks like we still haven't dropped the same check in query_api.rb, and we haven't even updated the gemspec to drop active record 3.2 support.  Until that is done, I'll leave this in for consistency.

arthurnn: make sense... I will drop it both. But after this is merged.
thanks

lib/identity_cache/parent_model_expiration.rb
dylanahsmith: Sure.  Are we trying to avoid try now?

lib/identity_cache/parent_model_expiration.rb
dylanahsmith: It is `parents_to_expire` is used as a set for duplication.

lib/identity_cache/parent_model_expiration.rb
```diff
@@ -4,44 +4,61 @@ module ParentModelExpiration # :nodoc:
...
         end
       end
 
-      true
+      cached_associations.each do |parent_class, only_on_foreign_key_change|
+        if new_parent && new_parent.is_a?(parent_class) && should_expire_identity_cache_parent?(foreign_key, only_on_foreign_key_change)
+          add_record_to_cache_expiry_set(parents_to_expire, new_parent)
+        end
+
+        if old_parent.try(:is_a?, parent_class)
```
arthurnn: same `try` comment applies here.

arthurnn:
LGTM! :shipit:

lib/identity_cache/parent_model_expiration.rb
BlakeMesdag: `is_a?` is defined on `Kernel` and is available all they down to `BasicObject`. When is this not defined?

lib/identity_cache/parent_model_expiration.rb
arthurnn: cool

PR #201, opened 2015-02-03
Remove undocumented overrides.

dylanahsmith:
cc @camilo 

## Problem

I hadn't noticed until seeing pull #171 that we allow all these internal configuration options to be overridden by the caller to `cache_belongs_to`, `cache_has_many` and `cache_has_one`.  I don't think these should be overridable, and I don't think we should even add them directly to the input options hash so that they aren't exposed at all.

I `grep`ed Shopify and couldn't find anywhere that we use these options.  It seems unlikely that others would use them.

## Solution

Duplicate the hash before any assignments using `.slice` to avoid making these options accessible on the input hash, and replace all those `||=` assignments with unconditional `=` assignments.

## Compatibility

These were undocumented, so I think it should be alright to remove these options without a major release, but it would break any code that used these options.

### DIFF
--- a/lib/identity_cache/belongs_to_caching.rb
+++ b/lib/identity_cache/belongs_to_caching.rb
@@ -11,13 +11,17 @@ module IdentityCache
       def cache_belongs_to(association, options = {})
         raise NotImplementedError if options[:embed]
 
+        unless association_reflection = reflect_on_association(association)
+          raise AssociationError, "Association named '#{association}' was not found on #{self.class}"
+        end
+
         options = {}
         self.cached_belongs_tos[association] = options
 
         options[:embed]                   = false
         options[:cached_accessor_name]    = "fetch_#{association}"
-        options[:foreign_key]             = reflect_on_association(association).foreign_key
-        options[:association_class]       = reflect_on_association(association).klass
+        options[:foreign_key]             = association_reflection.foreign_key
+        options[:association_class]       = association_reflection.klass
         options[:prepopulate_method_name] = "prepopulate_fetched_#{association}"
 
         build_normalized_belongs_to_cache(association, options)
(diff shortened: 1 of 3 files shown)


lib/identity_cache/belongs_to_caching.rb
```diff
@@ -9,18 +9,18 @@ module BelongsToCaching
 
     module ClassMethods
       def cache_belongs_to(association, options = {})
+        raise NotImplementedError if options[:embed]
+
+        options = {}
         self.cached_belongs_tos[association] = options
 
         options[:embed] ||= false
```
fbogsany: `options[:embed]` should be a direct assignment too.

dylanahsmith: Agreed.  Done.

TrangPham:
@dylanahsmith I also suggest that you check if the `reflect_on_association(association)` does not return nil. When it does the error returned is undefined method on nil. Could a better error be thrown? Something like `ActiveRecord::ConfigurationError: Association named 'fake' was not found on User; perhaps you misspelled it?`

dylanahsmith:
It looks like both cache_has_many and cache_has_one `raise InverseAssociationError unless self.reflect_on_association(association)` and that was just missing on cache_belongs_to, so I added a similar check.

I actually don't think `InverseAssociationError` is an appropriate name for the error in these cases though.  It seems like it only makes sense in `add_parent_expiry_hook` where it use using the `:inverse_name` option to find the inverse association.  As such, I changed the code to raise IdentityCache::AssociationError when not finding the inverse association.

In future we should add support for `:inverse_of` as an alternative association name for consistency with rails, and we should use the `:inverse_of` that is specified on the active record association to automatically detect these inverse associations.

TrangPham:
@dylanahsmith Thanks and agreed.

camilo:
is it also time to turn may of those into kwargs ?

dylanahsmith:
> is it also time to turn may of those into kwargs ?

Yes, we should do that.  However, that would definitely be a breaking change, and is outside the scope of this pull request.

dylanahsmith:
@fbogsany & @arthurnn ping for review

fbogsany:
:sheep:

PR #192, opened 2014-12-10
Fix namespace issue

RobWu:
Pull request for #191

PR #204, opened 2015-04-17
Fix namespace issue (again)

### LINKED ISSUE #191: Namespaced Classes not support anymore
Looks like the master branch has now problems with namespaces classes.

Worked with older versions.

Error is created on here: https://github.com/Shopify/identity_cache/blob/master/lib/identity_cache/configuration_dsl.rb#L280 

Would suggest to use ``` .parameterize.underscore ``` instead of ``` .underscore ```

```
/Users/jom/.rvm/gems/ruby-2.1.2@uddannelsestesten-2015/bundler/gems/identity_cache-c314bfcf65eb/lib/identity_cache/configuration_dsl.rb:282:in `class_eval': /Users/jom/.rvm/gems/ruby-2.1.2@uddannelsestesten-2015/bundler/gems/identity_cache-c314bfcf65eb/lib/identity_cache/configuration_dsl.rb:288: syntax error, unexpected '/', expecting ';' or '\n' (SyntaxError)
          def expire_parent_cache_b56_survey/question
                                             ^
/Users/jom/.rvm/gems/ruby-2.1.2@uddannelsestesten-2015/bundler/gems/identity_cache-c314bfcf65eb/lib/identity_cache/configuration_dsl.rb:290: syntax error, unexpected keyword_end, expecting end-of-input
	from /Users/jom/.rvm/gems/ruby-2.1.2@uddannelsestesten-2015/bundler/gems/identity_cache-c314bfcf65eb/lib/identity_cache/configuration_dsl.rb:282:in `add_parent_expiry_hook'
	from /Users/jom/.rvm/gems/ruby-2.1.2@uddannelsestesten-2015/bundler/gems/identity_cache-c314bfcf65eb/lib/identity_cache/configuration_dsl.rb:221:in `build_recursive_association_cache'
	from /Users/jom/.rvm/gems/ruby-2.1.2@uddannelsestesten-2015/bundler/gems/identity_cache-c314bfcf65eb/lib/identity_cache/configuration_dsl.rb:105:in `cache_has_many'
...
```

midu:
Hi,

I forked @RobWu's branch (https://github.com/Shopify/identity_cache/pull/192), rebased master into it, added a couple regression tests. Did a little method extraction (`def normalize_attribute_name`) that uses [@tjoyal's comment](https://github.com/Shopify/identity_cache/pull/192#issuecomment-69838165)

Fix #191 

Really hope we can get this fix out.

### DIFF
--- a/identity_cache.gemspec
+++ b/identity_cache.gemspec
@@ -31,6 +31,6 @@ Gem::Specification.new do |gem|
     gem.add_development_dependency('cityhash', '0.6.0')
     gem.add_development_dependency('mysql2')
     gem.add_development_dependency('pg')
-    gem.add_development_dependency('stackprof') if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("2.1.0")
+    gem.add_development_dependency('stackprof') if Gem::Version.new(RUBY_VERSION.dup) >= Gem::Version.new("2.1.0")
   end
 end
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -97,10 +97,12 @@ module IdentityCache
         options = options.slice(:embed, :inverse_name)
         options[:embed] = :ids unless options.has_key?(:embed)
         deprecate_embed_option(options, false, :ids)
-        options[:inverse_name] ||= self.name.underscore.to_sym
+
+        options[:inverse_name] ||= normalize_attribute_name(self.name)
         unless self.reflect_on_association(association)
           raise AssociationError, "Association named '#{association}' was not found on #{self.class}"
         end
+
         self.cached_has_manys[association] = options
 
         case options[:embed]
@@ -137,10 +139,12 @@ module IdentityCache
       def cache_has_one(association, options = {})
         options = options.slice(:embed, :inverse_name)
         options[:embed] = true unless options.has_key?(:embed)
-        options[:inverse_name] ||= self.name.underscore.to_sym
+
+        options[:inverse_name] ||= normalize_attribute_name(self.name)
         unless self.reflect_on_association(association)
           raise AssociationError, "Association named '#{association}' was not found on #{self.class}"
         end
+
         self.cached_has_ones[association] = options
 
         if options[:embed] == true
@@ -277,7 +281,7 @@ module IdentityCache
       end
 
       def after_action_name
-        @after_action_name ||= "expire_parent_cache_#{self.name.parameterize.underscore}"
+        @after_action_name ||= "expire_parent_cache_#{normalize_attribute_name(self.name)}"
       end
 
       def add_parent_expiry_hook(options)
@@ -310,6 +314,10 @@ module IdentityCache
           options[:embed] = new_value
           ActiveSupport::Deprecation.warn("`embed: #{old_value.inspect}` was renamed to `embed: #{new_value.inspect}` for clarity", caller(2))
         end
+      end
+
+      def normalize_attribute_name(attribute_name)
+        attribute_name.underscore.parameterize.underscore
       end
     end
   end
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -97,7 +97,7 @@ module IdentityCache
         options = options.slice(:embed, :inverse_name)
         options[:embed] = :ids unless options.has_key?(:embed)
         deprecate_embed_option(options, false, :ids)
-        options[:inverse_name] ||= normalize_attribute_name(self.name)
+        options[:inverse_name] ||= normalize_attribute_name_as_symbol(self.name)
         unless self.reflect_on_association(association)
           raise AssociationError, "Association named '#{association}' was not found on #{self.class}"
         end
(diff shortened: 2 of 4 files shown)


midu:
I extracted the normalization of the class name from all the places I could find `self.name.underscore`, I removed it in 502e1f1, let me know if you prefer to have them and I can revert to 0100eec

identity_cache.gemspec
```diff
@@ -31,6 +31,6 @@ Gem::Specification.new do |gem|
     gem.add_development_dependency('cityhash', '0.6.0')
     gem.add_development_dependency('mysql2')
     gem.add_development_dependency('pg')
-    gem.add_development_dependency('stackprof') if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("2.1.0")
+    gem.add_development_dependency('stackprof') if Gem::Version.new(RUBY_VERSION.dup) >= Gem::Version.new("2.1.0")
```
midu: `RUBY_VERSION` is frozen in (at least) ruby-1.9.3.p545, which prevents using the identity_cache gem

midu:
checking in: do you think these changes make sense? anything I can do to help it move forward?

pinging @arthurnn and @tjoyal since you both were active in the related PR #192

tjoyal:
Still good with the changes. :+1:

I do think there could still be some other issues on the same topic with the `options[:inverse_name]` used in that same file. But that can be tackled in another PR I suppose.

midu:
@tjoyal thanks for your comment. I am not exactly sure how `inverse_name` works, so I'm having a hard time writing a regression test for that. However, in this latest comment (https://github.com/midu/identity_cache/commit/80f2ea705d99c45d1ecda766d838011024e0cb58) I used the same normalization method for them. What do you think?

tjoyal:
> What do you think?

My point was more of a "this namespace issue could be affecting other part of the gem" but I'm all good at tackling them one at the time in subsequent PR.

As you can see with the test suite of your last commit, what should have been a no-brainer made the tests explode. (If you can figure it out that would be great)

midu:
@tjoyal right, the build broke on older rails because I returned a `String` instead of a `Symbol`, the last commit should fix that.

I couldn't find other places in the codebase that use `.underscore`, so I am hopping this is the only place that was affected by this specific issue.

camilo:
cc @dylanahsmith

midu:
any updates? anything I can do to help move this forward?

RobWu:
Would be great to get this merged in or any comment how to proceed.
Thanks

@dylanahsmith

identity_cache.gemspec
camilo: > is frozen in (at least) ruby-1.9.3.p545

Are you still running pre- 2.0. I've been thiking about dropping support for 1.9

lib/identity_cache/configuration_dsl.rb
```diff
@@ -309,6 +312,14 @@ def deprecate_embed_option(options, old_value, new_value)
           ActiveSupport::Deprecation.warn("`embed: #{old_value.inspect}` was renamed to `embed: #{new_value.inspect}` for clarity", caller(2))
         end
       end
+
+      def normalize_attribute_name(attribute_name)
+        attribute_name.underscore.parameterize.underscore
```
camilo: can you run the CPU bench after and before? this will get called a lot I think.

RobWu: You would prefer cache?
```ruby
def normalize_attribute_name(attribute_name)
  @normalize_attribute_name ||= {}
  @normalize_attribute_name[attribute_name] ||= attribute_name.underscore.parameterize.underscore
end
```

camilo: yes, well if parametrize does a lot of work.

identity_cache.gemspec
midu: I am running 1.9.3 on production. While the plan is to get to 2+ as soon as possible, it's not as trivial as it would seem. At least in my team's case.

My feeling is that `identity_cache` is quite a popular project and I wouldn't be surprised to learn that a lot of people still using it are running on 1.9, especially since it's a version still supported by the latest version of Rails.

I don't mean to tell you how to run your project of course, but since identity_cache is very tied to Rails, maybe it would make sense to have the ruby dependencies kind of follow Rails'. I see `Gemfile`s for versions going back to 3.2, which I don't believe supports ruby 2.

That being said, I am also all for moving forward, a pattern I've seen in some OSS projects is to lock a version for some older versions of ruby or rails and master has its own future forward dependencies.

My long way of saying that yes, I use 1.9, and I hope you'll still support it for a little longer :)

dylanahsmith:
Hasn't this been addressed by pull #199?  Although a regression test for the issue would still be appreciated.

midu:
@dylanahsmith I believe so, here's a PR https://github.com/Shopify/identity_cache/pull/210

identity_cache.gemspec
camilo: Fair, and the price is cheap for now. We might just release 1.0 and make it 4.x+ and 2.x+  only later on

midu: Cool, thanks! :)

I included that `dup` in #210.

midu:
closing this PR as the issue was fixed by #199 

PR #210 forks from master, adds the regression tests from this PR and the build is green

PR #207, opened 2015-05-07
Add hashing support with murmurhash3 and xxhash libraries

frobichaud:


arthurnn:
I am OK with this..
@camilo @dylanahsmith wdyt?

lib/identity_cache/cache_hash.rb
```diff
@@ -1,16 +1,23 @@
-# Use CityHash for fast hashing if it is available; use Digest::MD5 otherwise
-begin
-  require 'cityhash'
-rescue LoadError
+# Use a hashing library for fast hashing if it is available; use Digest::MD5 otherwise
+AVAILABLE_HASHING_GEMS = %w{xxhash murmurhash3 cityhash}
+
+hash_loaded = false
+AVAILABLE_HASHING_GEMS.each do |gem|
```
camilo: I can see a somewhat contrived scenario when this can be a problem assume you have cityhash in a project and for ...crazy reason you need to add xxhash, it would invalidate the caches without you knowing ?

frobichaud: @camilo should cityhash be the first option listed for backward compatibility?

camilo: I think so yes, that'd make it more stable

dylanahsmith: > crazy reason you need to add xxhash, it would invalidate the caches without you knowing ?

or more likely, you add some unrelated gem and it indirectly pulls in a dependancy like that.

frobichaud:
@camilo @dylanahsmith @arthurnn Fixed in latest commit

lib/identity_cache/cache_hash.rb
```diff
@@ -1,16 +1,23 @@
-# Use CityHash for fast hashing if it is available; use Digest::MD5 otherwise
-begin
-  require 'cityhash'
-rescue LoadError
+# Use a hashing library for fast hashing if it is available; use Digest::MD5 otherwise
+AVAILABLE_HASHING_GEMS = %w{cityhash xxhash murmurhash3}
```
camilo: This sill have the same problem tho, I mean is better because it uses the same as before but the fact that a xxhash user can unwittingly endup using cityhash remains.

I'm ok with the idea of this, but if we want it in the code base it should be more explicit.

camilo:
closing for now

RobWu:
Could we not just replace cityhash with xxhash?

cityhash is outdated (last release May 17, 2013), and not maintained anymore.. 
Does not install on my mac (10.11.3) nor on my ubuntu server (Ubuntu 14.04.4 LTS).

xxhash is still maintained & released (last release Marts 1, 2016)

PR #209, opened 2015-05-12
Stop testing on rails 3.2 and ruby 1.9.3 and 2.0.0

arthurnn:
@camilo @dylanahsmith WDYT?

It works right now, but those rails/ruby versions are super old, so I dont think we should keep supporting them.
If you are both OK, I will cut a new version before merging this, and we merge this moving forward.
(will also add a changelog)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,10 @@
 language: ruby
 
 rvm:
-  - 1.9.3
-  - 2.0.0
   - 2.1
   - 2.2
 
 gemfile:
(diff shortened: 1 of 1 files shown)


dylanahsmith:
Sure.  Were you wanting to take advantage of keyword arguments?

arthurnn:
> Were you wanting to take advantage of keyword arguments?

Thats one thing.

camilo:
:+1:

PR #217, opened 2015-07-31
Raise if a class method is called with a scope.

dylanahsmith:
@camilo & @arthurnn for review
cc @EmilS 

## Problem

Identity cache doesn't support rails scopes, but rails will automatically call class methods on relations, which IDC will ignore.  E.g. `user.associated_records.find_by_id(params[:record_id])` is a common pattern to prevent a user from records that it doesn't own, but `user.associated_records.find_by_id(params[:record_id])` will appear to work without proper testing for the case where record_id is for another user.

## Solution

When one of these class methods is called on an identity cache class method with a scope, then raise an exception.

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -20,6 +20,7 @@ module IdentityCache
       # Default fetcher added to the model on inclusion, it behaves like
       # ActiveRecord::Base.where(id: id).first
       def fetch_by_id(id)
+        raise_if_scoped
         return unless id
         raise NotImplementedError, "fetching needs the primary index enabled" unless primary_cache_index_enabled
         if IdentityCache.should_use_cache?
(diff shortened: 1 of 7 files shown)


lib/identity_cache/query_api.rb
```diff
@@ -77,6 +79,12 @@ def fetch_multi(*ids)
 
       private
 
+      def raise_if_scoped
+        if current_scope
```
camilo: nice, how stable is this API @arthurnn ?

camilo:
A+ @dylanahsmith can you add this to changelog as *breaking* this might :boom: your app.

lib/identity_cache/query_api.rb
dylanahsmith: It is undocumented, but was present from rails 3-2-stable to master.

dylanahsmith:
> can you add this to changelog

Done.

arthurnn:
:+1:

Issue #215, opened 2015-07-14
handle circular has_many, belongs_to associations


lrajlich:
models/asset.rb
```ruby
class Asset < ActiveRecord::Base
    include IdentityCache

    has_many        :custom_fields,            autosave: true
    cache_has_many  :custom_fields
end
```
models/custom_field.rb:
```ruby
class CustomField < ActiveRecord::Base
    include IdentityCache

    belongs_to :asset
    cache_belongs_to :asset
```

I need to require custom_field.rb in asset.rb otherwise I get this exception:
```
Uncaught exception: uninitialized constant Asset::CustomField
	/Users/lukerajlich/.rvm/gems/ruby-2.1.2/gems/activerecord-4.0.4/lib/active_record/inheritance.rb:129:in `compute_type'
	/Users/lukerajlich/.rvm/gems/ruby-2.1.2/gems/activerecord-4.0.4/lib/active_record/reflection.rb:178:in `klass'
	/Users/lukerajlich/.rvm/gems/ruby-2.1.2/bundler/gems/identity_cache-7a38bd64a5a5/lib/identity_cache/configuration_dsl.rb:234:in `build_id_embedded_has_many_cache'
	/Users/lukerajlich/.rvm/gems/ruby-2.1.2/bundler/gems/identity_cache-7a38bd64a5a5/lib/identity_cache/configuration_dsl.rb:110:in `cache_has_many'
	/Users/lukerajlich/Projects/dealpath/the_wall/models/asset.rb:7:in `<class:Asset>'
```

However, if I require custom_field
```
Uncaught exception: uninitialized constant CustomField::Asset
	/Users/lukerajlich/.rvm/gems/ruby-2.1.2/gems/activerecord-4.0.4/lib/active_record/inheritance.rb:129:in `compute_type'
	/Users/lukerajlich/.rvm/gems/ruby-2.1.2/gems/activerecord-4.0.4/lib/active_record/reflection.rb:178:in `klass'
	/Users/lukerajlich/.rvm/gems/ruby-2.1.2/bundler/gems/identity_cache-7a38bd64a5a5/lib/identity_cache/belongs_to_caching.rb:24:in `cache_belongs_to'
	/Users/lukerajlich/Projects/dealpath/the_wall/models/custom_field.rb:10:in `<class:CustomField>'
```

[...]

PR #218, opened 2015-08-03
postpone klass calls

lolownia:
This is a fix for issue #215 
identity_cache tries to call klass on associated models which are not yet loaded, raisng a missing constant exception.

### DIFF
--- a/lib/identity_cache/belongs_to_caching.rb
+++ b/lib/identity_cache/belongs_to_caching.rb
@@ -21,7 +21,7 @@ module IdentityCache
         options[:embed]                   = false
         options[:cached_accessor_name]    = "fetch_#{association}"
         options[:foreign_key]             = association_reflection.foreign_key
-        options[:association_class]       = association_reflection.klass
+        options[:association_class]       = association_reflection
         options[:prepopulate_method_name] = "prepopulate_fetched_#{association}"
 
         build_normalized_belongs_to_cache(association, options)
@@ -31,7 +31,7 @@ module IdentityCache
         self.class_eval(<<-CODE, __FILE__, __LINE__ + 1)
           def #{options[:cached_accessor_name]}
             if IdentityCache.should_use_cache? && #{options[:foreign_key]}.present? && !association(:#{association}).loaded?
-              self.#{association} = #{options[:association_class]}.fetch_by_id(#{options[:foreign_key]})
+              self.#{association} = #{options[:association_class].klass}.fetch_by_id(#{options[:foreign_key]})
             else
               #{association}
             end
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -209,7 +209,7 @@ module IdentityCache
       end
 
       def build_recursive_association_cache(association, options) #:nodoc:
-        options[:association_class]      = reflect_on_association(association).klass
+        options[:association_class]      = reflect_on_association(association)
         options[:cached_accessor_name]   = "fetch_#{association}"
         options[:records_variable_name]  = "cached_#{association}"
         options[:population_method_name] = "populate_#{association}_cache"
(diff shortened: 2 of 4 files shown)


lib/identity_cache/belongs_to_caching.rb
```diff
@@ -21,7 +21,7 @@ def cache_belongs_to(association, options = {})
         options[:embed]                   = false
         options[:cached_accessor_name]    = "fetch_#{association}"
         options[:foreign_key]             = association_reflection.foreign_key
-        options[:association_class]       = association_reflection.klass
+        options[:association_class]       = association_reflection
```
dylanahsmith: This should be renamed to `association_reflection` to avoid confusion.

dylanahsmith:
Cool.  Would you mind writing a regression test so we don't accidentally break this in the future?

lolownia:
Ouch!
I have just noticed that there this fix cannot work, because there is a call to ` add_parent_expiry_hook(options)` in configuration_dsl.rb:130 and in this function `klass` is needed.

I have used an ugly temporary workaround to `require_relative "my_associated_model"` in the beginning of model that caches associations....

The problem is logically you cannot use *loaded* classes (from klass call) in class_eval, because you cannot control the load order, and you can always run into situation when associated model is not yet loaded. So this really would need some fundamental change in identity_cache – perhaps delaying class resolution to later point, I don't know :<

lib/identity_cache/belongs_to_caching.rb
```diff
@@ -31,7 +31,7 @@ def build_normalized_belongs_to_cache(association, options)
         self.class_eval(<<-CODE, __FILE__, __LINE__ + 1)
           def #{options[:cached_accessor_name]}
             if IdentityCache.should_use_cache? && #{options[:foreign_key]}.present? && !association(:#{association}).loaded?
-              self.#{association} = #{options[:association_class]}.fetch_by_id(#{options[:foreign_key]})
+              self.#{association} = #{options[:association_class].klass}.fetch_by_id(#{options[:foreign_key]})
```
dylanahsmith: Use `class_name` instead of `klass` to avoid loading the class here.

lib/identity_cache/configuration_dsl.rb
```diff
@@ -279,7 +279,7 @@ def attribute_dynamic_fetcher(attribute, fields, values) #:nodoc:
       end
 
       def add_parent_expiry_hook(options)
-        child_class = options[:association_class]
+        child_class = options[:association_class].klass
```
lolownia: Thanks for comments, @dylanahsmith – do You have an idea how to deal with this one? here child_class must be known at class_eval time :/

dylanahsmith: I think we need to load the child class to set the after_commit callback that invalidates the cached associations.

lolownia: but if child is not yet loaded, this will not work.

dylanahsmith: rails will autoload the child class, but there shouldn't be a circular dependancy if you can get cache_belongs_to to not load the parent class.

lolownia: On Padrino with ActiveRecord the child class is not preloaded. I can fix this by adding `require "associatedmodel"` at the top of the file, but this is ugly. Maybe there's a way to alter this behaviour?

dylanahsmith: Adding `require "associatedmodel"` isn't as ugly as having to explicitly specify in the associated model that it needs to expire cached associations on the parent model.

lolownia: True but maybe there is a way to make identity_cache autoload these classes? How about putting `require options[:association_class].class_name.lower` or similar?

dylanahsmith: If autoloading is disabled, then that to me indicates that the developer using the library would prefer to manually require the dependant model.

dylanahsmith:
This was done by pull #226

PR #221, opened 2015-08-06
Raise when trying to cache a through association.

### LINKED ISSUE #220: Raise an exception when trying to cache a relation using :through option
We decided to not support the `:through` option in issue #56, but we don't make this clear to users of the library, leading to bugs that are caught later than necessary.

We should raise an exception when cache_has_one or cache_has_many is used on a relation with the `:through` option to avoid confusion.

dylanahsmith:
@camilo & @arthurnn for review
cc @RichardBlair

Fixes #220

### DIFF
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -308,6 +308,9 @@ module IdentityCache
         unless association_reflection = self.reflect_on_association(association)
           raise AssociationError, "Association named '#{association}' was not found on #{self.class}"
         end
+        if association_reflection.options[:through]
+          raise UnsupportedAssociationError, "caching through associations isn't supported"
+        end
         options[:inverse_name] ||= association_reflection.inverse_of.name if association_reflection.inverse_of
         options[:inverse_name] ||= self.name.underscore.to_sym
         child_class = association_reflection.klass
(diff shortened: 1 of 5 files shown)


lib/identity_cache/configuration_dsl.rb
```diff
@@ -98,9 +98,13 @@ def cache_has_many(association, options = {})
...
         deprecate_embed_option(options, false, :ids)
         options[:inverse_name] ||= self.name.underscore.to_sym
-        unless self.reflect_on_association(association)
+
+        unless association_reflection = self.reflect_on_association(association)
           raise AssociationError, "Association named '#{association}' was not found on #{self.class}"
         end
+        if association_reflection.options[:through]
+          raise UnsupportedAssociationError, "caching through associations isn't supported"
+        end
```
camilo: This code looks super similar to the one in line 145 could be DRYed up with a private method?

camilo:
one comment +1 otherwise

lib/identity_cache/configuration_dsl.rb
dylanahsmith: Yup.  Done.

camilo:
:+1:

arthurnn:
:shirt:

byroot:
:+1: I just stepped on this.

dylanahsmith:
I'll merge this after removing the cached through associations in Shopify.

PR #224, opened 2015-08-09
Return an array from fetch association methods to prevent chaining.

dylanahsmith:
@fbogsany & @tjoyal for review

## Problem

It was possible to chain the result of a fetched association (e.g. `record.fetch_association.where(important: true)` which could result in the association being loaded twice, once for fetching the association (e.g. `record.fetch_association`) and again to load the further scoped relation (e.g. the result of `.where(important: true)` in this case).

## Solution

Return an Array from the fetch association method rather than a relation to prevent chaining.  Any time they want a relation, they should use the association directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 #### 0.2.6 (unreleased)
 
 - Raise if a class method is called on a scope.  Previously the scope was ignored.
+- fetched has many associations returns an Array rather than a relation.
 
 #### 0.2.5
 
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -113,7 +113,7 @@ module IdentityCache
           association = reflection.association_class.new(record, reflection)
           association.target = coder_or_array.map {|e| record_from_coder(e) }
           association.target.each {|e| association.set_inverse_instance(e) }
-          association
+          association.target
         else
           record_from_coder(coder_or_array)
         end
@@ -125,7 +125,7 @@ module IdentityCache
         embedded_variable = record.instance_variable_get(:"@#{options[:records_variable_name]}")
         if IdentityCache.unmap_cached_nil_for(embedded_variable).nil?
           nil
-        elsif record.class.reflect_on_association(association).collection?
+        elsif embedded_variable.is_a?(Array)
           embedded_variable.map {|e| coder_from_record(e) }
         else
           coder_from_record(embedded_variable)
@@ -329,10 +329,9 @@ module IdentityCache
       ivar_full_name = :"@#{ivar_name}"
       if IdentityCache.should_use_cache?
         populate_recursively_cached_association(ivar_name, association_name)
-        assoc = IdentityCache.unmap_cached_nil_for(instance_variable_get(ivar_full_name))
-        assoc.is_a?(ActiveRecord::Associations::CollectionAssociation) ? assoc.reader : assoc
+        IdentityCache.unmap_cached_nil_for(instance_variable_get(ivar_full_name))
       else
-        send(association_name.to_sym)
+        association(association_name).load_target
       end
     end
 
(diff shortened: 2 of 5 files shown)


lib/identity_cache/query_api.rb
```diff
@@ -125,7 +125,7 @@ def get_embedded_association(record, association, options) #:nodoc:
         embedded_variable = record.instance_variable_get(:"@#{options[:records_variable_name]}")
         if IdentityCache.unmap_cached_nil_for(embedded_variable).nil?
           nil
-        elsif record.class.reflect_on_association(association).collection?
+        elsif embedded_variable.is_a?(Array)
           embedded_variable.map {|e| coder_from_record(e) }
         else
           coder_from_record(embedded_variable)
```
tjoyal: Correct me if I'm wrong (I'm no expert here), but IDC does not have the gem version in it's cache key.

By removing `record.class.reflect_on_association(association).collection?`, you'll end up with a `embedded_variable` that will be a `ActiveRecord::Associations::CollectionProxy` from existing cache before the change and `coder_from_record` will not like it.

test/denormalized_has_many_test.rb
```diff
@@ -88,4 +92,10 @@ def test_saving_associated_records_should_expire_itself_and_the_parents_cache
     IdentityCache.cache.expects(:delete).with(@record.primary_cache_index_key)
     child.save!
   end
+
+  def test_fetch_association_does_not_allow_chaining
```
tjoyal: I would prefer we test for the return type to be an `Array` than not to respond_to `where`.

Chaining is one of the things we want to prevent, but in the end what we really want is a raw array of objects not tied to AR.

tjoyal:
:+1: love how now the embed all vs embed ids behave the same way.

The only thing that concern me is the at deploy comment I put in there.

test/denormalized_has_many_test.rb
dylanahsmith: > I would prefer we test for the return type to be an `Array` than not to respond_to `where`.

The problem is that in rails 3.2 the collection proxy pretends to be an Array, so `record.associated_records.class` returns `Array` and `record.associated_records.class` returns `true`

lib/identity_cache/query_api.rb
dylanahsmith: We don't marshal associations, it is `set_embedded_association` that builds the association, which is used by `record_from_coder`.

tjoyal: :+1:

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 #### 0.2.6 (unreleased)
 
 - Raise if a class method is called on a scope.  Previously the scope was ignored.
+- fetched has many associations returns an Array rather than a relation.
```
rafaelfranca: It is a breaking change so I'd dump to 0.3 at least

rafaelfranca:
Yeah, it make sense to me.

rafaelfranca:
Anything missing on this PR?

tjoyal:
> The only thing that concern me is the at deploy comment I put in there.

That was already answered in another commit for me, so it is still a :+1: for me.

dylanahsmith:
> Anything missing on this PR?

Nope, I just need to fix places where we depend on a relation being returned from these fetch association methods.

camilo:
@dylanahsmith do we keep this one open or close it? I think is valuable so we can find someone to take over if you are busy.

dylanahsmith:
I would like to leave this open.  Anyone wanting to fix Shopify so it works with this change would be appreciated.  After that I think this PR just needs a rebase.

camilo:
wfm @dylanahsmith

daniellaniyo:
Should we close this PR now or are we doing it once we bump the version?

camilo:
Let's close now the objective has been achieved.

PR #226, opened 2015-08-13
Allow cache_belongs_to on polymorphic associations.

### LINKED ISSUE #215: handle circular has_many, belongs_to associations

### LINKED ISSUE #225: Polymorphic association caching
I have a `Patient` model that has a Single Table Inheritance with the `User` model. The User model has  polymorphic association with `PatientProfile` as profile.

Here is my `User` model :

```
class User < ActiveRecord::Base
  include IdentityCache

  # Include default devise modules. Others available are:
  # :lockable, :timeoutable and :omniauthable
  devise :database_authenticatable, :registerable, :confirmable,
         :recoverable, :rememberable, :trackable, :validatable

  belongs_to :profile, polymorphic: true, dependent: :destroy
end
```

Here is my `Patient` model :
```
class Patient < User
  delegate :other, to: :profile
end
```

And finally, the `PatientProfile` model : 
```
class PatientProfile < ActiveRecord::Base
  include IdentityCache

  has_one :user, as: :profile, dependent: :destroy
  cache_has_one :user, inverse_name: :profile, dependent: :destroy, embed: true
  accepts_nested_attributes_for :user
end
```

When I fetch my `User` all works perfectly : 
`[IdentityCache] (cache_backend) cache hit for IDC:5:blob:User:10236669911121066160:1`

Same for `Patient` :
`[IdentityCache] (cache_backend) cache hit for IDC:5:blob:User:10236669911121066160:1`

 And for `PatientProfile` :
`[IdentityCache] cache miss for IDC:5:blob:PatientProfile:15845076538005856020:1`

But when I try to fetch `PatientProfile` with the `Patient` model or the `User` model I got : 
```
[IdentityCache] (cache_backend) cache hit for IDC:5:blob:User:10236669911121066160:1
  PatientProfile Load (0.3ms)  SELECT  `patient_profiles`.* FROM `patient_profiles` WHERE `patient_profiles`.`id` = 1 LIMIT 1
```

I ran the command `User.fetch(1).profile` in the rails console.

Thanks a lot for the help

dylanahsmith:
@alexandcote does this fix #225 for you?
@lolownia & @lrajlich does this fix #215 for you?

## Problem

In cache_belongs_to we were assuming that a belongs_to association has a single association class that could be retrieved from the association reflection, but this isn't the case for polymorphic belongs_to associations, where a record is needed to get the actual class.

## Solution

Get the class from the association rather than the association reflection in the `fetch_#{association}` method.

This still needs some regression tests.

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -245,7 +245,7 @@ module IdentityCache
               end
 
               parent_record_to_child_records = Hash.new { |h, k| h[k] = [] }
-              child_records = details[:association_class].fetch_multi(*ids_to_parent_record.keys)
+              child_records = details[:association_reflection].klass.fetch_multi(*ids_to_parent_record.keys)
               child_records.each do |child_record|
                 parent_record = ids_to_parent_record[child_record.id]
                 parent_record_to_child_records[parent_record] << child_record
@@ -262,11 +262,16 @@ module IdentityCache
             if details[:embed] == true
               raise ArgumentError.new("Embedded belongs_to associations do not support prefetching yet.")
             else
+              reflection = details[:association_reflection]
+              if reflection.options[:polymorphic]
+                raise ArgumentError.new("Polymorphic belongs_to associations do not support prefetching yet.")
+              end
+
               ids_to_child_record = records.each_with_object({}) do |child_record, hash|
-                parent_id = child_record.send(details[:foreign_key])
+                parent_id = child_record.send(reflection.foreign_key)
                 hash[parent_id] = child_record if parent_id.present?
               end
-              parent_records = details[:association_class].fetch_multi(ids_to_child_record.keys)
+              parent_records = reflection.klass.fetch_multi(ids_to_child_record.keys)
               parent_records.each do |parent_record|
                 child_record = ids_to_child_record[parent_record.id]
                 child_record.send(details[:prepopulate_method_name], parent_record)
@@ -288,8 +293,8 @@ module IdentityCache
             raise ArgumentError.new("Unknown cached association #{association} listed for prefetching")
           end
 
-          if details && details[:association_class].respond_to?(:prefetch_associations, true)
-            details[:association_class].send(:prefetch_associations, sub_associations, next_level_records)
+          if details && details[:association_reflection].klass.respond_to?(:prefetch_associations, true)
+            details[:association_reflection].klass.send(:prefetch_associations, sub_associations, next_level_records)
           end
         end
       end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Raise when trying to cache a through association. Previously it wouldn't be invalidated properly.
 - Raise if a class method is called on a scope.  Previously the scope was ignored.
 - Raise if a class method is called on a subclass of one that included IdentityCache. This never worked properly.
+- Fix cache_belongs_to on polymorphic assocations.
 
 #### 0.2.5
 
--- a/README.md
+++ b/README.md
@@ -125,7 +125,9 @@ IdentityCache tries to figure out both sides of an association whenever it can s
 
 ``` ruby
 class Metafield < ActiveRecord::Base
+  include IdentityCache
   belongs_to :owner, :polymorphic => true
+  cache_belongs_to :owner
 end
 
 class Product < ActiveRecord::Base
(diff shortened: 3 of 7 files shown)


lib/identity_cache/query_api.rb
```diff
@@ -259,11 +259,16 @@ def prefetch_associations(associations, records)
             if details[:embed] == true
               raise ArgumentError.new("Embedded belongs_to associations do not support prefetching yet.")
             else
+              reflection = details[:association_reflection]
+              if reflection.options[:polymorphic]
+                raise ArgumentError.new("Polymorphic belongs_to associations do not support prefetching yet.")
```
dylanahsmith: This is for simplicity, since when dealing with multiple records, we might have a mix of types to fetch.  It could be implemented by grouping by type to fetch them, but I'm not sure if we want to support that complexity at the moment.

alexandcote:
Yeah ! This work perfectly for me ! Thanks a lot @dylanahsmith for this fix

lolownia:
Unfortunately no, sorry. Did not fix #215 
Please check out a minimal example here:
https://github.com/lolownia/identity_cache_bug
so far I can workaround this bug by setting autoload in config/initializers like:
```
autoload :Foo, File.expand_path('../../../models/foo.rb', __FILE__)
```

dylanahsmith:
> so far I can workaround this bug by setting autoload in config/initializers

issue #215 is about a circular references, not about expecting identity_cache to implement autoloading for you, so it sounds like this would fix that issue.

dylanahsmith:
Updated the documentation to include the example usage of a polymorphic cache_belongs_to

I added a regression test for issue #225.  I'm not sure what is a good way to test issue #215, so there is no regression test for that.

dylanahsmith:
@byroot & @fbogsany would either of you mind giving this a review?

dylanahsmith:
> I'm not sure what is a good way to test issue #215, so there is no regression test for that.

I guess a test explicitly for this isn't needed, since calling `reflect_on_association(association).klass` in cache_belongs_to would fail on a polymorphic association, as is tested in test_fetching_polymorphic_belongs_to_association

byroot:
I can't right now but îm flaging et t

byroot:
:clap: LGTM

rafaelfranca:
:shipit:

PR #227, opened 2015-08-18
There is no try

rafaelfranca:
The return of `where` is always an `ActiveRecord::Relation`, so there is no need for using try here if it is always present and always respond to `first`.


rafaelfranca:
Review @byroot @dylanahsmith

dylanahsmith:
:+1:

byroot:
:ship:

PR #228, opened 2015-08-25
Prevent encode_with from overriding :class

byroot:
Custom `encode_with` shouldn't be able to mess with reserved IDC attributes.

@rafaelfranca for review please.


rafaelfranca:
Make sense. :shipit:

PR #229, opened 2015-09-04
Raise if id embedded associated class hasn't included IdentityCache

### LINKED ISSUE #203: cache_has_many only work with :embed => true
I'm running rails 4.2.1 and I've setup identity_cache following the docs every thing working as expected until I tried to remove :embed => true or set it to false

I always get this output from rails console when run modle_instance.fetch_* 

```
NoMethodError: undefined method `fetch_multi' for #<Class:0x007f24099829a0>
from /home/ams/.rbenv/versions/2.1.4/lib/ruby/gems/2.1.0/gems/activerecord-4.2.1/lib/active_record/dynamic_matchers.rb:26:in `method_missing'
```

I even tried to downgrade rails to 4.1.9 and I get the same error . any help on that matter ? 

Thank you

dylanahsmith:
@byroot for review

Fixes #203

## Problem

Changing the the `embed` option on a cache_has_many from `embed: true` to `embed: :ids` can result in missing method errors, which can be confusing.  This can happen when the associated class hasn't included IdentityCache, which is needed to fetch the associated record when only the id is embedded.

## Solution

Raise in cache_has_many with a clear exception message so the problem can quickly be identified and resolved.

### DIFF
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -315,6 +315,9 @@ module IdentityCache
         options[:inverse_name] ||= self.name.underscore.to_sym
         child_class = association_reflection.klass
         raise InverseAssociationError unless child_class.reflect_on_association(options[:inverse_name])
+        unless options[:embed] == true || child_class.include?(IdentityCache)
+          raise UnsupportedAssociationError, "associated class #{child_class} must include IdentityCache to be cached without full embedding"
+        end
       end
     end
   end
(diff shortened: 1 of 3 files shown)


lib/identity_cache/configuration_dsl.rb
```diff
@@ -315,6 +315,9 @@ def ensure_cacheable_association(association, options)
         options[:inverse_name] ||= self.name.underscore.to_sym
         child_class = association_reflection.klass
         raise InverseAssociationError unless child_class.reflect_on_association(options[:inverse_name])
+        unless options[:embed] == true || child_class.include?(IdentityCache)
+          raise UnsupportedAssociationError, "associated class must include IdentityCache to be cached without full embedding"
```
byroot: Maybe include the name of the associated class in the error message? For extra clarity.

byroot:
One nitpick. Otherwise LGTM.

lib/identity_cache/configuration_dsl.rb
dylanahsmith: Makes sense.  Done.

PR #230, opened 2015-09-04
Raise when calling a class method on a subclass.

### LINKED ISSUE #182: Including identity cache in a base model that other models inherit from causes issues
Ran into the issue where if a model inherited a base model that included Identity Cache the secondary indexes would be shared across models generating undefined method errors. The offending line was line 15 of configuration dsl which create the cache_indexes array on the class that includes identity cache (ModelBase in this case).

https://github.com/Shopify/identity_cache/blob/master/lib/identity_cache/configuration_dsl.rb#L15
````ruby
class ModelBase < ActiveRecord::Base
  include IdentityCache
end

class User < ModelBase
    cache_index :user, :key
end

class Store < ModelBase
    cache_index :store, :id
end
````

I resolved the error by including Identity Cache individually on all models and not on a base model.

### PR DESCRIPTION
**Pull request by dylanahsmith**
@camilo & @byroot for review

Fixes #182

## Problem

When IdentityCache is included in the base class for a model that uses single-table inheritance, then calling `.fetch` on a subclass could result in us caching `nil` when the record exists when the subclass doesn't match the `type` of the record.

E.g. if we have the following models

```
class Company < ActiveRecord::Base
  include IdentityCache
end
class Firm < Company; end
class Client < Company; end
```

and row id 1 is for type Client, then calling `Firm.fetch(1)` will cache `nil` for id 1, resulting in `Client.fetch(1)` to not find the record that actually exists.

Additionally, caching associations on a derived class actually caches the association on the base class.

We never intended to support calling IDC class methods on derived classes, so it would be best to have them fail hard rather than in subtle ways from being misused.

## Solution

Store the model that IdentityCache is included on, and check the model when IdentityCache class methods are called.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Raise when trying to cache a through association. Previously it wouldn't be invalidated properly.
 - Raise if a class method is called on a scope.  Previously the scope was ignored.
+- Raise if a class method is called on a subclass of one that included IdentityCache. This never worked properly.
 
 #### 0.2.5
 
--- a/lib/identity_cache.rb
+++ b/lib/identity_cache.rb
@@ -29,6 +29,7 @@ module IdentityCache
   end
   class UnsupportedScopeError < StandardError; end
   class UnsupportedAssociationError < StandardError; end
+  class DerivedModelError < StandardError; end
 
   class << self
     include IdentityCache::CacheHash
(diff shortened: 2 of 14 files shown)


byroot:
LGTM. Even thought actually supporting it would be better. But I have no idea if it's even possible.

camilo:
:+1: adding support would be nice but for now at least add some guardrails.

dylanahsmith:
We could support it in the future, but it won't be a trivial change.  For now I just want to make anything that isn't supported fail fast rather than introduce bugs.

waheedel:
Supporting is possible. I have a branch but can't create a PR.

PR #231, opened 2015-09-08
Avoid using mysql2 0.4 in CI which doesn't work released versions of rails

dylanahsmith:
@rafaelfranca for review

[CI is failing on master](https://travis-ci.org/Shopify/identity_cache/builds/79304602) because it is trying to use mysql2 version 0.4 which was released yesterday, but [active record only loads mysql2 0.3.x](https://github.com/rails/rails/blob/v4.2.4/activerecord/lib/active_record/connection_adapters/mysql2_adapter.rb#L3)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,2 +1,4 @@
 source 'https://rubygems.org'
 gemspec
+
+gem 'mysql2', '~> 0.3.13'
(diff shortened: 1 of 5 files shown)


rafaelfranca:
:shipit:

PR #233, opened 2015-09-17
Test with Rails 4.2

byroot:
I think CI is broken for 4.2. We should make it green.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,3 @@ sudo: false
 before_script:
   - mysql -e 'create database identity_cache_test'
   - psql -c 'create database identity_cache_test;' -U postgres
-
-matrix:
-  allow_failures:
-    - gemfile: Gemfile.rails42


byroot:
Ok, so the failing test is because of the new way attributes are serialized, which we actually monkey patched internally.


@rafaelfranca I think we should implement the serialization method in IDC directly as we talked about a couple times. That would fix this test.

cc @dylanahsmith

dylanahsmith:
I've started working on this in the [use-instantiate](https://github.com/shopify/identity_cache/compare/use-instantiate) branch.  I decided to use the active record instantiate method to create a record, which is what is used when a record is loaded from the database. There are still test failures with postgres, because I used the attributes before type cast which caused some differences in what gets serialized for postgres.

rafaelfranca:
O_O. I thought we had tests green with 4.2.

byroot:
If you want to be retro compatible here's what you need:

```ruby
#dump:
attributes.send(:initialized_attributes).transform_values(&:value_for_database)

#load
self.class.attributes_builder.build_from_database(coder['attributes'])
```

rafaelfranca:
We could try `instantiate` but @byroot and I think it would be better if we define a serialization protocol in IDC itself so it will not break between Rails versions.

Let's get @sgrif involved since he did almost all the work in Rails type casting. @sgrif What do you thing about using `instantiate` to load a AR object from memcache?

dylanahsmith:
> We could try `instantiate` but @byroot and I think it would be better if we define a serialization protocol in IDC itself so it will not break between Rails versions.

That is why I wanted to use `instantiate` because it is given the data as it comes from the database, exactly what we want to cache.  I didn't want to depend on coders, which included a bunch of ruby objects, and I don't think we should serialize the data after type casting, because that means we are caching a transformation that might changes across versions of the code.

sgrif:
Sorry, I missed this ping. Yes, `instantiate` is the right thing to use here.

PR #234, opened 2015-09-24
Don't store IDC loaded records in the belongs to association.

dylanahsmith:
@byroot & @rafaelfranca for review

## Problem

IdentityCache tries to make a point of being opt-in.  As such, accessing the association directly shouldn't return records loaded from IdentityCache.

However, fetching a `cache_belongs_to` association with IdentityCache will store the result in the underlying `belongs_to` association.

## Solution

Use the `@cached_#{association}` instance variable to store the result of fetching a cache_belongs_to association, the same way we do it for cache_has_many and cache_has_one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Raise if a class method is called on a scope.  Previously the scope was ignored.
 - Raise if a class method is called on a subclass of one that included IdentityCache. This never worked properly.
 - Fix cache_belongs_to on polymorphic assocations.
+- Fetching a cache_belongs_to association no longer loads the belongs_to association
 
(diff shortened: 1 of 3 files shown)


byroot:
Quick question: What should be the behavior if the record is reloaded?

dylanahsmith:
The [CacheInvalidation](https://github.com/Shopify/identity_cache/blob/3304e2d193cc321de884a704fbac2e6a6feb6548/lib/identity_cache/cache_invalidation.rb) module clears these instance variables on reload.

byroot:
Perfect. LGTM then. :shipit:

rafaelfranca:
:shipit:

PR #235, opened 2015-10-06
Get the build passing on 4.2

sgrif:
The YAML format used by Active Record has changed as of Rails 4.2. A few
tests would not pass no matter what, as the objects being serialized
didn't implement equality comparison. I've fixed this upstream in
https://github.com/rails/rails/commit/9e25e0e1735f2ccca69679243aa8cf0885104164
and
https://github.com/rails/rails/commit/cfb371e96a2789790359ec1db779df490fd253d6.
I've also backported those fixes for our tests.

I'm actually wondering whether we need to bump the serialization
version. Is the contract that we never change the format that things are
serialized? Or is the contract that we accept anything serialized with
an older version? Rails 4.2 will accept the Rails 4.1 yaml as of
https://github.com/rails/rails/commit/438662ab32715d3b780fb54b0743cc2e19843086

That change is unreleased however, so maybe it's worth bumping it
anyway, just so we can be compatible with 4.2 as quickly as possible.

It's been brought up that we want to switch to a serialization format
that is independent of the Active Record version. I think this is a good
idea, and I'm going to explore it. However, I wanted to push up the
minimum possible change set to get us working w/ 4.2 first.

Fixes #233 

/cc @byroot @rafaelfranca


byroot:
> Is the contract that we never change the format that things are serialized?

We've always namespaced the IDC keys with the Rails version. But it's a big PITA because it means we need to delete twice when we upgrade Rails.

That is why a format independent from the Rails version would be so nice.

Another very nice thing would be to embed the serialization version in the payload, so that the deserializer can be backward compatible. This would tremendously reduce the IDC induced headaches.

> The YAML format used by Active Record has changed as of Rails 4.2.

Yes, and it's about 2 to 3 times the size of the previous format, which is not really acceptable, at least not for us.

I think we should go directly to an AR independent, versioned format.

fbogsany:
> That is why a format independent from the Rails version would be so nice.

It would also be nice to have a format independent from the Ruby version. The test for serialization compatibility was intended as a canary for Marshal format changes.

sgrif:
> That is why a format independent from the Rails version would be so nice.

Yes, I agree. Like I said I just wanted to get the minimal set of changes to be 4.2 compatible out the door first, as that's potentially a little bit more involved.

dylanahsmith:
Just changing the serialization format fixes the tests for rails 4.2, but causes it to fail with all previous versions.

> It's been brought up that we want to switch to a serialization format
that is independent of the Active Record version. I think this is a good
idea, and I'm going to explore it. However, I wanted to push up the
minimum possible change set to get us working w/ 4.2 first.

See https://github.com/shopify/identity_cache/compare/use-instantiate which does what you are suggesting.  I think it just needs the serialized record fixture to be split by adapter, since making it independent of rails meant making it dependant on what the database adapter returns for the query.

sgrif:
Ah good point. Yeah, we definitely need to go independent right away. I had something slightly different in mind for going fully independent, we're discussing in the rails room in Slack.

dylanahsmith:
> Another very nice thing would be to embed the serialization version in the payload, so that the deserializer can be backward compatible. This would tremendously reduce the IDC induced headaches.

:+1: as long as we don't need to A/B test between the two serialization versions, since that would cause contention from the two versions treating what the other version wrote as a miss, so we would keep refilling the cache.  This is fine for all the schema related versioning.

PR #238, opened 2015-10-29
Fix N+1 from fetching embedded ids on a cache miss.

dylanahsmith:
@rafaelfranca & @sgrif for review

## Problem

On a cache miss, the embedded associations are preloaded using `.includes`, but the id embedded associations are just preloaded by loading the ids for each parent record.

E.g. If a Product model has `cache_has_many :images, embed: :ids`, then multiple products are loaded with `Product.fetch_multi(*ids)`, then `product.image_ids` will get called on each product, causing a query per product.

## Solution

Query the associated model using a where condition on the foreign key, and selecting both the primary key and foreign key in the query, so the list of ids can be grouped by the parent record's id.

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -182,24 +182,24 @@ module IdentityCache
 
       def preload_id_embedded_associations(records)
         return if records.empty?
-        each_id_embedded_association do |name, options|
+        each_id_embedded_association do |options|
           reflection = options.fetch(:association_reflection)
           child_model = reflection.klass
           scope = child_model.all
           scope = scope.instance_exec(nil, &reflection.scope) if reflection.scope
 
           pairs = scope.where(reflection.foreign_key => records.map(&:id)).pluck(reflection.foreign_key, reflection.active_record_primary_key)
-          ids_by_parent = {}
+          ids_by_parent = Hash.new{ |hash, key| hash[key] = [] }
           pairs.each do |parent_id, child_id|
-            (ids_by_parent[parent_id] ||= []) << child_id
+            ids_by_parent[parent_id] << child_id
           end
 
           records.each do |parent|
-            child_ids = ids_by_parent[parent.id] || []
+            child_ids = ids_by_parent[parent.id]
             parent.instance_variable_set(:"@#{options.fetch(:ids_variable_name)}", child_ids)
           end
         end
-        recursively_embedded_associations.each do |name, options|
+        recursively_embedded_associations.each_value do |options|
           child_model = options.fetch(:association_reflection).klass
           if child_model.respond_to?(:preload_id_embedded_associations)
             child_records = records.flat_map(&options.fetch(:cached_accessor_name).to_sym)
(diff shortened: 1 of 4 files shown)


lib/identity_cache/query_api.rb
```diff
@@ -179,9 +175,43 @@ def require_if_necessary #:nodoc:
...
+        each_id_embedded_association do |name, options|
+          reflection = options.fetch(:association_reflection)
+          child_model = reflection.klass
+          scope = child_model.all
+          scope = scope.instance_exec(nil, &reflection.scope) if reflection.scope
+
+          pairs = scope.where(reflection.foreign_key => records.map(&:id)).pluck(reflection.foreign_key, reflection.active_record_primary_key)
+          ids_by_parent = {}
+          pairs.each do |parent_id, child_id|
+            (ids_by_parent[parent_id] ||= []) << child_id
```
rafaelfranca: Maybe defining the default value to hash with `[]` would make this method simpler?

lib/identity_cache/query_api.rb
```diff
@@ -179,9 +175,43 @@ def require_if_necessary #:nodoc:
...
+          end
+
+          records.each do |parent|
+            child_ids = ids_by_parent[parent.id] || []
```
rafaelfranca: The default value in the hash would also make this `||` unnecessary.

lib/identity_cache/query_api.rb
```diff
@@ -179,9 +175,43 @@ def require_if_necessary #:nodoc:
...
-        object.send(:populate_association_caches) if object
-        object
+        record = self.includes(cache_fetch_includes).reorder(nil).where(primary_key => id).first
+        preload_id_embedded_associations([record]) if record
+        record
+      end
+
+      def preload_id_embedded_associations(records)
+        return if records.empty?
+        each_id_embedded_association do |name, options|
```
rafaelfranca: Is `name` being used here? Does it make sense to pass it if it is not being used?

lib/identity_cache/query_api.rb
```diff
@@ -179,9 +175,43 @@ def require_if_necessary #:nodoc:
...
+          pairs.each do |parent_id, child_id|
+            (ids_by_parent[parent_id] ||= []) << child_id
+          end
+
+          records.each do |parent|
+            child_ids = ids_by_parent[parent.id] || []
+            parent.instance_variable_set(:"@#{options.fetch(:ids_variable_name)}", child_ids)
+          end
+        end
+        recursively_embedded_associations.each do |name, options|
```
rafaelfranca: `name` is not being used here too.

dylanahsmith: switched to use `each_value` here

lib/identity_cache/query_api.rb
dylanahsmith: Sure, I added a default proc to the hash, so now this line is just `ids_by_parent[parent_id] << child_id`

lib/identity_cache/query_api.rb
dylanahsmith: Ok.  It just passes the association options now.

rafaelfranca:
:shipit:

PR #239, opened 2015-10-29
Remove code for rails 3.2 compatibility.

dylanahsmith:
@sgrif & @rafaelfranca for review

## Problem

Pull request #209 stopped testing rails 3.2 in CI, because we wanted to stop supporting that old a version of rails.  However, we never bumped the minimum activerecord gem version.  Since then, the test suite has stopped passing with rails 3.2, so let's make this lack of support for that version of rails more explicit.

## Solution

* Bumped the activerecord version in the gemspec
* Removed the gemfile for testing against rails 3.2
* Removed some code for active record versions lower than 4.0.4, where after_commit wasn't being called for touching a record.

### DIFF
--- a/Gemfile.rails32
+++ b/Gemfile.rails32
@@ -1,6 +1,0 @@
-source 'https://rubygems.org'
-gemspec
-
-gem 'activerecord', '~> 3.2.16'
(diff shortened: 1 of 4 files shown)


rafaelfranca:
LGTM.

PR #240, opened 2015-10-29
Test Ruby 2.2.3

fbogsany:
Temporary PR to test compatibility with ruby 2.2.3. I don't intend to merge this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: ruby
 
 rvm:
   - 2.1
-  - 2.2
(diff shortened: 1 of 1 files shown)


rafaelfranca:
I think it is worth to merge it. Using 2.2 we avoid to changing that every time a new Ruby version is released but we have to wait travis upgrade their default infrastructure that can take some time. In this case 2.2.3 was released 2 months ago and 2.2 is still 2.2.0.

PR #241, opened 2015-10-29
Fix exec_query expectation test failures on rails 4.2 with postgres.

dylanahsmith:
@byroot & @rafaelfranca for review

## Problem

When running the tests with rails 4.2 and postgres, there are test failures from some `connection.expects(:exec_query)` expectations.  E.g.

```
  1) Failure:
FetchTest#test_fetch_with_garbage_input [/home/travis/.rvm/gems/ruby-2.1.5/gems/activerecord-4.2.4/lib/active_record/connection_adapters/abstract/database_statements.rb:351]:
unexpected invocation: #<ActiveRecord::ConnectionAdapters::PostgreSQLAdapter:0x2438f08>.exec_query('SELECT  \'items\'.* FROM \'items\' WHERE \'items\'.\'id\' = $1 LIMIT 1', 'Item Load', [[#<ActiveRecord::ConnectionAdapters::PostgreSQLColumn:0x26f0318>, 'garbage']])
unsatisfied expectations:
- expected exactly once, not yet invoked: #<ActiveRecord::ConnectionAdapters::PostgreSQLAdapter:0x2438f08>.exec_query('SELECT  \'items\'.* FROM \'items\' WHERE \'items\'.\'id\' = 0 LIMIT 1', any_parameters)
```

These tests seem quite fragile, by testing internal method calls in rails.  They are also intrusive, by stubbing out those methods and returning a fake value, rather than performing a integrated test of the actual behaviour of dependant components.

## Solution

Setup datastores into the right state so mocked values don't need to be returned to test different code paths, and use `assert_queries` to make sure the right number of queries are made to the database.

### DIFF
--- a/test/attribute_cache_test.rb
+++ b/test/attribute_cache_test.rb
@@ -13,85 +13,76 @@ class AttributeCacheTest < IdentityCache::TestCase
     IdentityCache.cache.clear
   end
 
-  def test_attribute_values_are_returned_on_cache_hits
-    IdentityCache.cache.expects(:fetch).with(@name_attribute_key).returns('foo')
-    assert_equal 'foo', AssociatedRecord.fetch_name_by_id(1)
-  end
-
   def test_attribute_values_are_fetched_and_returned_on_cache_misses
     fetch = Spy.on(IdentityCache.cache, :fetch).and_call_through
-    expects_fetch_associated_record_name_by_id(1, returns: 'foo')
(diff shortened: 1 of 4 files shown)


byroot:
Makes perfect sense to me. :+1:

rafaelfranca:
:+1: :heart: integration tests.

PR #242, opened 2015-10-29
Use ActiveRecord instantiate method to create a record from the cache.

dylanahsmith:
Fixes #233 

@sgrif & @byroot for review

## Problem

Rails 4.2 changed the implementation of encode_with, such that it returns some active record objects, and gets marshal dumped to about 3 times the size.  We have worked around this in Shopify using a monkey patch, but we should instead serialize records in a rails independent way.

## Solution

Use `attributes_before_type_cast` for serializing records and `instantiate` for deserializing records, since `instantiate` is the same path used to create records from the results of the query as returned form the database adapter.  As such, the format is dependant on the database client, instead of being dependant on rails, so shouldn't cause format changes from upgrading rails.

Previously we were storing type casted values, so I bumped the serialization format, and cleaned up the coder hash (e.g. attributes now uses a symbol rather than string key, like the other keys).


sgrif:
:shipit:

byroot:
:clap:

rafaelfranca:
:heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #243, opened 2015-11-03
Fix a usage of respond_to? for a private method.

dylanahsmith:
@rafaelfranca for review

Pull #238 introduced this check for `child_model.respond_to?(:preload_id_embedded_associations)`, but `preload_id_embedded_associations` is a private method, so it always returned `false`.  This wasn't caught by the tests, since it would just result it the same N+1 it was trying to fix, just in the case of fetching ids for deeply associated records.

I used `child_model.include?(IdentityCache)` instead, since I think that check it is a less fragile check than changing it to `child_model.respond_to?(:preload_id_embedded_associations, true)`.  I also added a regression test for this case.


rafaelfranca:
:shipit:

PR #245, opened 2015-11-04
Compact nils when preloading id embedded associations

dylanahsmith:
@rafaelfranca for review

## Problem

I got `NoMethodError: undefined method 'fetch_association' for nil:NilClass` like errors when testing identity_cache master on Shopify.  This can happen when preloading an id embedded association through a embedded has_one association, since `records.flat_map(&options.fetch(:cached_accessor_name).to_sym)` was returning `[nil]` when the has_one association is `nil`.

## Solution

Compact the array and add a regression test.


rafaelfranca:
:shipit:

PR #247, opened 2015-12-03
[fetch_multi] Only call cache fetcher once per unique key

wkjagt:
When calling `#fetch_multi` with an array that contains duplicates, these duplicates are passed all the way into the cache backend. This PR only sends the unique values.

What do you think @gmalette ?

### DIFF
--- a/test/fetch_multi_test.rb
+++ b/test/fetch_multi_test.rb
@@ -121,6 +121,18 @@ class FetchMultiTest < IdentityCache::TestCase
     assert_equal [@joe, @bob, @joe], Item.fetch_multi(@joe.id, @bob.id, @joe.id)
   end
 
+  def test_fetch_multi_with_duplicate_ids_hits_backend_once_per_id
+    cache_response = {
+      @joe_blob_key => cache_response_for(@joe),
+      @bob_blob_key => cache_response_for(@bob),
+    }
+
+    fetcher.expects(:fetch_multi).with([@joe_blob_key, @bob_blob_key]).returns(cache_response)
+    result = Item.fetch_multi(@joe.id, @bob.id, @joe.id)
+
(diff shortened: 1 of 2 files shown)


test/fetch_multi_test.rb
```diff
@@ -121,6 +121,18 @@ def test_fetch_multi_duplicate_ids
...
+  def test_fetch_multi_with_duplicate_ids_hits_backend_once_per_id
+    cache_response = {
+      @joe_blob_key => cache_response_for(@joe),
+      @bob_blob_key => cache_response_for(@bob),
+    }
+
+    fetcher.expects(:fetch_multi).with([@joe_blob_key, @bob_blob_key]).returns(cache_response)
+    result = Item.fetch_multi(@joe.id, @bob.id, @joe.id)
+
+    assert_equal [@joe, @bob, @joe], result
```
gmalette: I don't understand the magic at play here... where are these mapped against the keys?

wkjagt: The actual magic is in [`QueryApi#fetch_multi`](https://github.com/Shopify/identity_cache/blob/fetch-multi/one-query-per-unique-key/lib/identity_cache/query_api.rb#L48)

gmalette: Ah gotcha, it's not the same `fetch_multi` method

gmalette:
I think this is fine, please make sure to ping people that have better knowledge on this

wkjagt:
@dylanahsmith @rafaelfranca not sure who to ping on this, but you seem to be quite active in here. Would you mind taking a look?

dylanahsmith:
LGTM

rafaelfranca:
LGTM

wkjagt:
Thanks!

PR #248, opened 2016-01-14
Fixed an issue that is unique cache_index keys are returned as array

parangdae:
https://github.com/Shopify/identity_cache/blob/master/lib/identity_cache/fallback_fetcher.rb#L34-L41

key = "hello"
result = @cache_backend.read(key)
expected result was "world" but returned ["world"]

'dalli', '~> 2.7.5'
'identity_cache', '~> 0.2.5'
'cityhash', '~> 0.8.1'

### DIFF
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -199,7 +199,7 @@ module IdentityCache
         cache_key = rails_cache_index_key_for_fields_and_values(fields, values)
         id = IdentityCache.fetch(cache_key) { identity_cache_conditions(fields, values).limit(1).pluck(primary_key).first }
         unless id.nil?
-          record = fetch_by_id(id)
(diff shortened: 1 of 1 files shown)


lib/identity_cache/configuration_dsl.rb
```diff
@@ -199,7 +199,7 @@ def identity_cache_single_value_dynamic_fetcher(fields, values) # :nodoc:
         cache_key = rails_cache_index_key_for_fields_and_values(fields, values)
         id = IdentityCache.fetch(cache_key) { identity_cache_conditions(fields, values).limit(1).pluck(primary_key).first }
```
dylanahsmith: This would always store just an id in the cache.  Did you encounter an Array from adding a `unique: true` option to an existing cache_index?

dylanahsmith:
Please add a regression test, since it isn't obvious how to reproduce this error.

parangdae:
Sorry, my mistakes. it isn't now reproduce.

PR #249, opened 2016-01-15
Add support for includes option on fetch methods

dylanahsmith:
@rafaelfranca & @richardmonette for review

## Problem

When there is a need to fetch a deeply associated belongs to association, then we can't rely on embedding to avoid an N+1, since cache_belongs_to doesn't support embedding.  However, `fetch_multi` supports an `:includes` option to batch the fetching of that association for multiple parent records, e.g. `Item.fetch_multi(ids, include: { associated_records: :deeply_associated_record })` where `:deeply_associated_record` is a belongs_to association.

The problem is that the `:includes` option is only support on fetch_multi, but not `fetch`, `fetch_by_id` or the cache index generated `fetch_by_`* methods.  This is inconsistent and could lead to ugly workarounds where `fetch_multi` is used where `fetch` or `fetch_by_id` are more appropriate.  Or where `cache_attribute :id, by:` is used instead of `cache_index` to get access to the `id`, since `cache_index` doesn't provide access to the `id` in its `fetch_by_`* methods and uses `fetch_by_id` internally.

## Solution

Support the `:includes` option on the above mentioned fetch methods.

I noticed that `fetch_by_id` wasn't preloading all the embedded associations when `IdentityCache.should_use_cache?` returns `false` (i.e. during a database transaction).  I needed to change this inconsistency because `prefetch_association` assumes all embedded associations have already been loaded.

## Future Work

I noticed that `cache_attribute` could be used to implement `cache_index`, if it had a concept of a non-unique index.  This would also provide `fetch_id_by_`* methods for cache_index, which would make it more flexible. **Edit: https://github.com/Shopify/identity_cache/pull/250 does this**

It might also be useful to expose `prefetch_associations` for fetching associations on a record that was already loaded.  Although, then `prefetch_associations` could be called on a record that doesn't have all its embedded associations loaded, which is an assumption it can currently make.

### DIFF
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -52,19 +52,19 @@ module IdentityCache
 
         if options[:unique]
           self.instance_eval(ruby = <<-CODE, __FILE__, __LINE__ + 1)
-            def fetch_by_#{field_list}(#{arg_list})
-              identity_cache_single_value_dynamic_fetcher(#{fields.inspect}, [#{arg_list}])
+            def fetch_by_#{field_list}(#{arg_list}, options={})
+              identity_cache_single_value_dynamic_fetcher(#{fields.inspect}, [#{arg_list}], options)
             end
 
             # exception throwing variant
-            def fetch_by_#{field_list}!(#{arg_list})
-              fetch_by_#{field_list}(#{arg_list}) or raise ActiveRecord::RecordNotFound
+            def fetch_by_#{field_list}!(#{arg_list}, options={})
+              fetch_by_#{field_list}(#{arg_list}, options) or raise ActiveRecord::RecordNotFound
             end
           CODE
         else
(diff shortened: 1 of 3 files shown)


rafaelfranca:
LGTM

richardmonette:
:+1:

richardmonette:
Just re-reading this now and experimenting with its use in Shopify core, did we miss this spot? https://github.com/Shopify/identity_cache/blob/master/lib/identity_cache/configuration_dsl.rb#L254

dylanahsmith:
I'm not sure if it is needed there, since we want to do the prefetching on the top level query, otherwise it is a nested query that could be performed multiple times.  I was also hesitant about adding it to a memoized `fetch_#{associations}` method, since that would complicate the memoization by having to also check for the preloads.

However, I did mention that

> It might also be useful to expose prefetch_associations for fetching associations on a record that was already loaded.

which seems more reasonable than adding support for `:includes` on those memoized methods.

PR #250, opened 2016-01-16
Implement cache_index using cache_attribute.

dylanahsmith:
@rafaelfranca, @camilo for review

## Problem

`cache_index *columns, unique: true` cache the same thing as `cache_attribute :id, by: columns` so we should just use cache_attribute to implement cache_index to simplify the codebase.

## Solution

This PR adds a `:unique` option to cache_attribute, because that is an option `cache_index` supports and `cache_attribute` used to assume the index was always unique.

`cache_index` uses the `cache_attribute` by calling the `fetch_id_by_`* for the index to get the record ids, then uses `fetch_by_id` or `fetch_multi` to get the record itself.  This means there are new `fetch_id_by_`* methods defined for each cache_index.

The cache key for cache_attribute now changes when the `:unique` option changes to fix #248.

### DIFF
--- a/lib/identity_cache/cache_key_generation.rb
+++ b/lib/identity_cache/cache_key_generation.rb
@@ -35,12 +35,12 @@ module IdentityCache
         "#{rails_cache_key_namespace}blob:#{base_class.name}:#{rails_cache_key_prefix}:"
       end
 
-      def rails_cache_index_key_for_fields_and_values(fields, values)
-        "#{rails_cache_key_namespace}index:#{base_class.name}:#{rails_cache_string_for_fields_and_values(fields, values)}"
-      end
-
-      def rails_cache_key_for_attribute_and_fields_and_values(attribute, fields, values)
-        "#{rails_cache_key_namespace}attribute:#{base_class.name}:#{attribute}:#{rails_cache_string_for_fields_and_values(fields, values)}"
+      def rails_cache_key_for_attribute_and_fields_and_values(attribute, fields, values, unique)
+        "#{rails_cache_key_namespace}" \
+          "attr#{unique ? '' : 's' }" \
+          ":#{base_class.name}" \
+          ":#{attribute}" \
+          ":#{rails_cache_string_for_fields_and_values(fields, values)}"
       end
 
       def rails_cache_key_namespace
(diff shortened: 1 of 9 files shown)


lib/identity_cache/configuration_dsl.rb
```diff
@@ -42,18 +42,19 @@ module ClassMethods
       # * unique: if the index would only have unique values
       #
       def cache_index(*fields)
-        ensure_base_model
         raise NotImplementedError, "Cache indexes need an enabled primary index" unless primary_cache_index_enabled
         options = fields.extract_options!
-        self.cache_indexes.push fields
+        unique = options[:unique] || false
```
rafaelfranca: maybe fetch?

rafaelfranca: ah, got it, it can't be `nil` otherwise will default to `true` in `cache_attribute`

lib/identity_cache/query_api.rb
```diff
@@ -375,39 +375,23 @@ def expire_primary_index # :nodoc:
...
-            old_cache_index_key = secondary_cache_index_key_for_previous_values(fields)
-            IdentityCache.cache.delete(old_cache_index_key) unless old_cache_index_key == new_cache_index_key
-          end
-        end
-      end
-    end
-
     def expire_attribute_indexes # :nodoc:
-      cache_attributes.try(:each) do |(attribute, fields)|
+      cache_attributes.try(:each) do |(attribute, fields, unique)|
```
rafaelfranca: Can we remove this `each`? It seems to be always an array and we are expecting it to be an array at `cache_attribute` so I think there is no point of not expecting here.

rafaelfranca:
LGTM. Way simpler indeed.

lib/identity_cache/cache_key_generation.rb
```diff
@@ -35,12 +35,12 @@ def prefixed_rails_cache_key
...
 
-      def rails_cache_index_key_for_fields_and_values(fields, values)
-        "#{rails_cache_key_namespace}index:#{base_class.name}:#{rails_cache_string_for_fields_and_values(fields, values)}"
-      end
-
-      def rails_cache_key_for_attribute_and_fields_and_values(attribute, fields, values)
-        "#{rails_cache_key_namespace}attribute:#{base_class.name}:#{attribute}:#{rails_cache_string_for_fields_and_values(fields, values)}"
+      def rails_cache_key_for_attribute_and_fields_and_values(attribute, fields, values, unique)
+        "#{rails_cache_key_namespace}" \
+          "attr#{unique ? '' : 's' }" \
```
camilo: nitpick: I'd put the `unique ? '' : 's' ` outside of the interpolation for better readability.

camilo:
:+1:

lib/identity_cache/cache_key_generation.rb
dylanahsmith: sure, done

lib/identity_cache/query_api.rb
dylanahsmith: Yes, I'm not sure why we did that in the first place.  Done.

lib/identity_cache/query_api.rb
```diff
@@ -375,39 +375,23 @@ def expire_primary_index # :nodoc:
...
-          if !was_new_record?
-            old_cache_index_key = secondary_cache_index_key_for_previous_values(fields)
-            IdentityCache.cache.delete(old_cache_index_key) unless old_cache_index_key == new_cache_index_key
-          end
-        end
-      end
-    end
-
     def expire_attribute_indexes # :nodoc:
-      cache_attributes.try(:each) do |(attribute, fields)|
```
dylanahsmith: Looks like a name conflict with a deprecated rails method might be why this was added (https://github.com/rails/rails/blob/v4.2.5/activerecord/lib/active_record/attribute_methods/read.rb#L39)

lib/identity_cache/configuration_dsl.rb
```diff
@@ -42,18 +40,19 @@ module ClassMethods
...
+        cache_attribute primary_key, by: fields, unique: unique
```
dylanahsmith: @camilo this is the line that causes problems for Shopify's container build, which doesn't allow schema queries, since it happens during initialization.

PR #251, opened 2016-01-18
Rename cache_attributes to cache_indexes to avoid method name conflict.

dylanahsmith:
@rafaelfranca for review

When testing with pull #250 I noticed ```DEPRECATION WARNING: Calling `cache_attributes` is no longer necessary.``` warnings. This is because active record defines a deprecated class method named `cache_attributes`, and identity cache has a class_attribute of the same name to store the options given to the `cache_attribute` DSL method.

This naming conflict somehow resulted in problems with IDC finding the inverse name of associations.  I'm not sure why this was a symptom of the problem, but fixing the naming conflict resolved the issue.  I'm also not sure why this problem didn't turn in up the IDC test suite.

I chose to rename the `cache_attributes` class attribute to `cache_indexes`.  The naming is still appropriate, since it is basically an index used to look up an attribute.

### DIFF
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -4,7 +4,7 @@ module IdentityCache
 
     included do |base|
       base.class_attribute :cached_model
-      base.class_attribute :cache_attributes
+      base.class_attribute :cache_indexes
       base.class_attribute :cached_has_manys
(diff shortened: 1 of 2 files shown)


rafaelfranca:
Oh. That is weird. Yeah, renaming is safer here. :shipit:

PR #252, opened 2016-01-19
Allow prefetch_associations to be used on existing records.

dylanahsmith:

## Problem

I would like to be able to prefetch associations on records which have already been loaded.

An example of where this can be useful is when performing an update on a record.  The record needs to be loaded from the database to avoid propagating cache corruption, but that prevents prefetching unloaded associations using the existing record when creating an HTTP response that corresponds to that update.

This can also be useful for normalized associations, like when calling `fetch_#{association}` on a `cache_belongs_to` association or on an id embedded association, which doesn't support the `:includes` option, since it would complicate memorization.

## Solution

I made the `prefetch_associations` method public and made it work with records not loaded from IDC.  Embedded associations are loaded by fetching the owner of the association from the cache, then moving the loaded association records onto the records passed to `prefetch_associations`.

### DIFF
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -201,17 +201,14 @@ module IdentityCache
         options[:cached_accessor_name]   = "fetch_#{association}"
         options[:records_variable_name]  = "cached_#{association}"
 
+        self.class_eval(<<-CODE, __FILE__, __LINE__ + 1)
+          def #{options[:cached_accessor_name]}
+            fetch_recursively_cached_association('#{options[:records_variable_name]}', :#{association})
+          end
+        CODE
 
-        unless instance_methods.include?(options[:cached_accessor_name].to_sym)
-          self.class_eval(<<-CODE, __FILE__, __LINE__ + 1)
-            def #{options[:cached_accessor_name]}
-              fetch_recursively_cached_association('#{options[:records_variable_name]}', :#{association})
-            end
-          CODE
-
-          options[:only_on_foreign_key_change] = false
-          add_parent_expiry_hook(options)
-        end
+        options[:only_on_foreign_key_change] = false
+        add_parent_expiry_hook(options)
       end
 
       def build_id_embedded_has_many_cache(association, options) #:nodoc:
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -275,16 +275,15 @@ module IdentityCache
 
       def fetch_embedded_associations(records)
         associations = embedded_associations
-        return true if associations.empty?
+        return if associations.empty?
 
-        return false unless primary_cache_index_enabled
+        return unless primary_cache_index_enabled
 
-        records_by_id = records.index_by(&:id)
-        cached_records = fetch_multi(records.map(&:id))
+        cached_records_by_id = fetch_multi(records.map(&:id)).index_by(&:id)
 
         associations.each_value do |options|
-          cached_records.each do |cached_record|
-            record = records_by_id.fetch(cached_record.id)
+          records.each do |record|
+            next unless cached_record = cached_records_by_id[record.id]
             if options[:embed] == :ids
               cached_association = cached_record.public_send(options.fetch(:cached_ids_name))
               record.instance_variable_set(:"@#{options.fetch(:ids_variable_name)}", cached_association)
@@ -294,7 +293,6 @@ module IdentityCache
             end
           end
         end
-        true
       end
 
       def prefetch_embedded_association(records, association, details)
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -297,6 +297,9 @@ module IdentityCache
       end
 
       def prefetch_embedded_association(records, association, details)
+        # Make the same assumption as ActiveRecord::Associations::Preloader, which is
+        # that all the records have the same associations loaded, so we can just check
+        # the first record to see if an association is loaded.
         first_record = records.first
         return if first_record.association(association).loaded?
         iv_name_key = details[:embed] == true ? :records_variable_name : :ids_variable_name
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -79,6 +79,7 @@ module IdentityCache
 
         case associations
         when nil
+          # do nothing
         when Symbol
           prefetch_one_association(associations, records)
         when Array
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -74,6 +74,32 @@ module IdentityCache
         records
       end
 
+      def prefetch_associations(associations, records)
+        return if records.empty?
+
+        case associations
+        when nil
+        when Symbol
+          prefetch_one_association(associations, records)
+        when Array
+          associations.each do |association|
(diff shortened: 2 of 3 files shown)


lib/identity_cache/query_api.rb
```diff
@@ -246,96 +272,102 @@ def find_batch(ids)
...
 
-              parent_record_to_child_records = Hash.new { |h, k| h[k] = [] }
-              child_records = details[:association_reflection].klass.fetch_multi(*ids_to_parent_record.keys)
-              child_records.each do |child_record|
-                parent_record = ids_to_parent_record[child_record.id]
-                parent_record_to_child_records[parent_record] << child_record
-              end
+      def prefetch_embedded_association(records, association, details)
+        first_record = records.first
+        return if first_record.association(association).loaded?
```
dylanahsmith: This is the same assumption that active record makes in `ActiveRecord::Associations::Preloader`

lib/identity_cache/configuration_dsl.rb
```diff
@@ -201,17 +201,14 @@ def build_recursive_association_cache(association, options) #:nodoc:
         options[:cached_accessor_name]   = "fetch_#{association}"
         options[:records_variable_name]  = "cached_#{association}"
 
+        self.class_eval(<<-CODE, __FILE__, __LINE__ + 1)
+          def #{options[:cached_accessor_name]}
+            fetch_recursively_cached_association('#{options[:records_variable_name]}', :#{association})
+          end
+        CODE
 
-        unless instance_methods.include?(options[:cached_accessor_name].to_sym)
```
eapache: this change doesn't seem strictly related?

dylanahsmith: Right, this was a useless check, but I could make a separate PR for it.

boourns:
LGTM

lib/identity_cache/configuration_dsl.rb
eapache: No need, I just thought I was missing something.

lib/identity_cache/query_api.rb
```diff
@@ -246,96 +272,102 @@ def find_batch(ids)
...
+        iv_name_key = details[:embed] == true ? :records_variable_name : :ids_variable_name
+        return if first_record.instance_variable_get(:"@#{details[iv_name_key]}")
+        fetch_embedded_associations(records)
+      end
 
-              parent_record_to_child_records.each do |parent_record, child_records|
-                parent_record.send(details[:prepopulate_method_name], child_records)
+      def prefetch_one_association(association, records)
```
eapache: this method is enormous... can we split the three cases into their own methods for readability?

lib/identity_cache/query_api.rb
```diff
@@ -74,6 +74,32 @@ def fetch_multi(*ids)
         records
       end
 
+      def prefetch_associations(associations, records)
+        return if records.empty?
+
+        case associations
+        when nil
```
eapache: this is an intentional do-nothing case? would be nice to have a comment making that explicit, otherwise people might think it was supposed to fall-through

dylanahsmith: Yup, we supported that previously with the `when nil` in hashify_includes_structure.  I'll add a comment.

lib/identity_cache/query_api.rb
eapache: put this as an in-code comment please

lib/identity_cache/query_api.rb
```diff
@@ -246,96 +272,102 @@ def find_batch(ids)
         ids.map{ |id| records_by_id[id] }
       end
 
-      def prefetch_associations(associations, records)
-        associations = hashify_includes_structure(associations)
+      def fetch_embedded_associations(records)
+        associations = embedded_associations
+        return true if associations.empty?
```
eapache: AFAICT nothing is using these return values?

lib/identity_cache/query_api.rb
```diff
@@ -246,96 +272,102 @@ def find_batch(ids)
...
 
-        associations.each do |association, sub_associations|
-          case
-          when details = cached_has_manys[association]
+        return false unless primary_cache_index_enabled
 
-            if details[:embed] == true
-              child_records = records.map(&details[:cached_accessor_name].to_sym).flatten
+        records_by_id = records.index_by(&:id)
+        cached_records = fetch_multi(records.map(&:id))
```
eapache: not sure this line makes sense in the context of the change - isn't the whole point that if we were passed in a complete record already (e.g. one fetched straight from the DB in an update scenario) that we *don't* want to go to the cache for that record because it could have incorrect associations, and we just want to go the cache for the associations of the "concrete" record we already have?

lib/identity_cache/query_api.rb
eapache: Oh, I see when reviewing with `?w=0` that this method actually didn't really change, just got unindented

lib/identity_cache/query_api.rb
dylanahsmith: The whole point of this method *is* to go the cache for the associations.  However, those associations are embedded on a record, so we need to fetch the record to get at the associations.

I don't know what you mean about *incorrect associations*.  In relation to a mutation, the cache would have been invalidated to avoid fetching anything incorrect.

eapache: > The whole point of this method is to go the cache for the associations.

Yes.

> However, those associations are embedded on a record, so we need to fetch the record to get at the associations.

We already have the record; that's the point of this PR, to load associations when we already have th record.

dylanahsmith: I don't really understand what is the problem with this line then, it seems essential to this pull request.

lib/identity_cache/query_api.rb
```diff
@@ -246,96 +272,102 @@ def find_batch(ids)
...
+        return true if associations.empty?
 
-        associations.each do |association, sub_associations|
-          case
-          when details = cached_has_manys[association]
+        return false unless primary_cache_index_enabled
 
-            if details[:embed] == true
-              child_records = records.map(&details[:cached_accessor_name].to_sym).flatten
+        records_by_id = records.index_by(&:id)
```
dylanahsmith: Actually, this assumes `records` are unique, which could be avoided by indexing the cached records.

lib/identity_cache/query_api.rb
dylanahsmith: Oh right.  I wasn't quite sure how to handle the case where the parent record has been deleted and can't be refetched.  In many cases the associations have also been deleted, but it is probably safer to fallback to the database for this edge case.  We probably don't need to worry about efficiency for this edge case, so I never ended up using the result.

I'll change these to not explicitly return anything.

lib/identity_cache/query_api.rb
dylanahsmith: `# do nothing` was added

lib/identity_cache/query_api.rb
eapache: We already have the record(s). Why are we re-fetching the record(s)?

dylanahsmith: We have the records, but not their embedded associations.  How can you fetch the embedded associations from the cache?  The only place they are stored in the cache is in the same cache entry that would store these records that we already have.

eapache:
LGTM

PR #253, opened 2016-01-20
Lazily resolve primary key.

dylanahsmith:
@camilo for review

## Problem

cache_index is now making calls to `primary_key` when it is called, rather than lazily when the fetch_* method it defines is called.  However, we have removed dependance on the schema during initialization in Shopify, and now calling `primary_key` would cause the container build to fail.

## Solution

We can follow the naming convention in active record, where `id` is used in the method name to refer to the primary key, even when the primary key actually has a different name.  Similarly, we can use `id` in the cache key, even when the primary key has a different name.  As such, I moved the implementation of cache_attribute into a `cache_attribute_by_alias` that takes an additional alias_name parameter to use a hard coded `id` alias for the `cache_index` attribute.

### DIFF
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -43,7 +43,7 @@ module IdentityCache
         raise NotImplementedError, "Cache indexes need an enabled primary index" unless primary_cache_index_enabled
         options = fields.extract_options!
         unique = options[:unique] || false
-        cache_attribute primary_key, by: fields, unique: unique
+        cache_attribute_by_alias('primary_key', 'id', by: fields, unique: unique)
 
         field_list = fields.join("_and_")
         arg_list = (0...fields.size).collect { |i| "arg#{i}" }.join(',')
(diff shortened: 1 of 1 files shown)


lib/identity_cache/configuration_dsl.rb
```diff
@@ -171,31 +171,35 @@ def cache_has_one(association, options = {})
...
+      end
+
+      def disable_primary_cache_index
+        ensure_base_model
+        self.primary_cache_index_enabled = false
+      end
+
+      private
+
+      def cache_attribute_by_alias(attribute, alias_name, options)
```
camilo: why is the `alias_name` a param, can it ever change? and be someting diff from id ?

dylanahsmith: `cache_index` always passes `id`, but `cache_attribute` passes the attribute name.

lib/identity_cache/configuration_dsl.rb
```diff
@@ -171,31 +171,35 @@ def cache_has_one(association, options = {})
       # * by: Other attribute or attributes in the model to keep values indexed. Default is :id
       # * unique: if the index would only have unique values. Default is true
       def cache_attribute(attribute, options = {})
+        cache_attribute_by_alias(attribute.inspect, attribute, options)
```
camilo: why inspect?

camilo:
one small q :+1:

lib/identity_cache/configuration_dsl.rb
dylanahsmith: Because I needed something that would be lazily evaluated. It is used in the `instance_eval` string with `attribute_dynamic_fetcher(#{attribute}`.

Issue #255, opened 2016-02-02
Allow should_use_cache? to be overridden on a model


brettbevers:
The application that I'm working on uses database transactions to wrap actions that need to be cleanly rolled back in case of an error. There is a lot of static reference data that is retrieved within the transaction. I am using IdentityCache to cache this same reference data throughout the application, but the cache is intentionally not used when there is an open transaction. It would be ideal if I could configure this particular behavior per model. I've forked this repo and made some changed there that I believe accomplishes this (https://github.com/brettbevers/identity_cache/tree/allow-override-should-use-cache).  Can someone please look over my commit and point out any potential problems?

juancaldera:
+1

theturtle32:
:+1: +1

dylanahsmith:
The idea sounds good to me.  Open up a pull request with your commit to make it easy to comment on the code.

brettbevers:
Thanks @dylanahsmith !

Sirupsen:
Fixed in https://github.com/Shopify/identity_cache/pull/291#issue-173330279

PR #256, opened 2016-02-02
Allow IdentityCache.should_use_cache? to be overridden per model

brettbevers:
Changes referenced in issue https://github.com/Shopify/identity_cache/issues/255

### DIFF
--- a/lib/identity_cache.rb
+++ b/lib/identity_cache.rb
@@ -13,6 +13,7 @@ require "identity_cache/cache_hash"
 require "identity_cache/cache_invalidation"
 require "identity_cache/cache_fetcher"
 require "identity_cache/fallback_fetcher"
+require "identity_cache/configuration"
 
 module IdentityCache
   CACHED_NIL = :idc_cached_nil
@@ -43,6 +44,7 @@ module IdentityCache
     def included(base) #:nodoc:
       raise AlreadyIncludedError if base.include?(IdentityCache::ConfigurationDSL)
 
+      base.send(:include, IdentityCache::Configuration)
       base.send(:include, ArTransactionChanges) unless base.include?(ArTransactionChanges)
       base.send(:include, IdentityCache::BelongsToCaching)
       base.send(:include, IdentityCache::CacheKeyGeneration)
@@ -89,8 +91,8 @@ module IdentityCache
     # == Parameters
     # +key+ A cache key string
     #
-    def fetch(key)
-      if should_use_cache?
+    def fetch(key, use_cache=should_use_cache?)
+      if use_cache
         unmap_cached_nil_for(cache.fetch(key) { map_cached_nil_for yield })
       else
         yield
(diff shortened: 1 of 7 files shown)


eddify:
:+1:

jasonbosco:
+1

brettbevers:
@dylanahsmith Is there any interest in this PR?

lib/identity_cache.rb
```diff
@@ -89,8 +91,8 @@ def should_use_cache? # :nodoc:
     # == Parameters
     # +key+ A cache key string
     #
-    def fetch(key)
-      if should_use_cache?
+    def fetch(key, use_cache=should_use_cache?)
```
dylanahsmith: This is inconsistent with fetch_multi which uses a keyword argument.  A keyword argument is probably appropriate here too

test/attribute_cache_test.rb
```diff
@@ -86,6 +86,22 @@ def test_fetching_by_attribute_delegates_to_block_if_transactions_are_open
     end
   end
 
+  def test_overriding_should_use_cache_when_fetching_by_attribute
+    IdentityCache.cache.expects(:fetch).at_least_once
```
dylanahsmith: I don't like how `assert_queries(0)` depends `IdentityCache.fetch` being mocked out.  I think you can just warm the cache with `AssociatedRecord.fetch_name_by_id(1)` to avoid mocking

lib/identity_cache.rb
```diff
@@ -89,8 +91,8 @@ def should_use_cache? # :nodoc:
     # == Parameters
     # +key+ A cache key string
     #
-    def fetch(key)
-      if should_use_cache?
+    def fetch(key, use_cache: should_use_cache?)
```
camilo: I'm not sure I like this particular implementation, I think giving the client of the model the responsibility/power to bypass the cache a call at the time will end up in a lot of cognitive overhead on both implementers and reviewers. 

From reading the original issue it seems that you have context dependent needs to disable the cache, as in "I really want to make sure  that under this context model `M` does not try to use the cache".

What about a block syntax?

```
M.disable_cacheing do
end
```

or

```
IdentityCache.disable_cacheing_for(*M) do
end
```

Granted that would cause some surprise on the methods  called inside the  block, but would make it easier to separate contexts when you are reasoning about what should be cached and what should not.

brettbevers: Actually, my use case has to do with forcing the use of the cache for particular models in a circumstance in which the cache would otherwise be automatically bypassed. In particular, IdentityCache#fetch currently assumes that it should never use the cache if there is an open database transaction. I want to replace this assumption with logic that is specific to each model and, consequently, specific to each call to `fetch`.

Sirupsen:
Fixed in https://github.com/Shopify/identity_cache/pull/291#issue-173330279

PR #259, opened 2016-02-19
Raise an exception when trying to cache an association scoped with a join

dylanahsmith:
@Thibaut & @richardmonette for review

## Problem

I caught an attempt to cache an association that was scoped with a join in Shopify, which wouldn't have been properly invalidated if the joined table changed.  We should check for this when attempting to cache these associations and raise an exception so that we don't catch this problem in production, which would be hard to debug.

## Solution

Apply the scope to the associated class, then check to see if the relation has any `joins_values`

### DIFF
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -291,9 +291,13 @@ module IdentityCache
         if association_reflection.options[:through]
           raise UnsupportedAssociationError, "caching through associations isn't supported"
         end
+        child_class = association_reflection.klass
+        scope = association_reflection.scope
+        if scope && !child_class.all.instance_exec(&scope).joins_values.empty?
+          raise UnsupportedAssociationError, "caching association scoped with a join isn't supported"
+        end
         options[:inverse_name] ||= association_reflection.inverse_of.name if association_reflection.inverse_of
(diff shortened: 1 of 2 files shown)


rafaelfranca:
LGTM

lib/identity_cache/configuration_dsl.rb
```diff
@@ -291,9 +291,13 @@ def ensure_cacheable_association(association, options)
         if association_reflection.options[:through]
           raise UnsupportedAssociationError, "caching through associations isn't supported"
         end
+        child_class = association_reflection.klass
+        scope = association_reflection.scope
+        if scope && !child_class.instance_exec(&scope).joins_values.empty?
```
Thibaut: nit: `if scope && child_class.instance_exec(&scope).joins_values.present?`

Thibaut:
:+1:

richardmonette:
:+1:

lib/identity_cache/configuration_dsl.rb
dylanahsmith: I prefer to use `.empty?` rather than `.blank?` or `.present?` unless I want to handle the receiver being `nil`.

dylanahsmith:
CI passes in Shopify with this branch after I changed the `.instance_exec(&scope)` to be on `child_class.all` instead of `child_class`

PR #260, opened 2016-02-22
Delay association scope check out of initialization.

dylanahsmith:
@richardmonette & @Thibaut for review
cc @fw42

## Problem

https://github.com/Shopify/identity_cache/pull/259 broke Shopify production container builds since it uses a dummy database configuration and active record makes schema queries when building a relation.  That means we can't really check the cached association when it is defined.

## Solution

I moved the check into the cache key generation, since that is done just before reading or writing to the cache.

### DIFF
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -291,13 +291,9 @@ module IdentityCache
         if association_reflection.options[:through]
           raise UnsupportedAssociationError, "caching through associations isn't supported"
         end
-        child_class = association_reflection.klass
-        scope = association_reflection.scope
-        if scope && !child_class.all.instance_exec(&scope).joins_values.empty?
-          raise UnsupportedAssociationError, "caching association scoped with a join isn't supported"
-        end
         options[:inverse_name] ||= association_reflection.inverse_of.name if association_reflection.inverse_of
         options[:inverse_name] ||= self.name.underscore.to_sym
+        child_class = association_reflection.klass
         raise InverseAssociationError unless child_class.reflect_on_association(options[:inverse_name])
         unless options[:embed] == true || child_class.include?(IdentityCache)
           raise UnsupportedAssociationError, "associated class #{child_class} must include IdentityCache to be cached without full embedding"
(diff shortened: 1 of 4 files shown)


test/denormalized_has_many_test.rb
```diff
@@ -94,24 +94,29 @@ def test_saving_associated_records_should_expire_itself_and_the_parents_cache
     child.save!
   end
 
-  def test_unsupported_through_assocation
-    assert_raises IdentityCache::UnsupportedAssociationError, "caching through associations isn't supported" do
-      Item.has_many :deeply_through_associated_records, :through => :associated_records, foreign_key: 'associated_record_id', inverse_of: :item, :class_name => 'DeeplyAssociatedRecord'
-      Item.cache_has_many :deeply_through_associated_records, :embed => true
+  class CheckAssociationTest < IdentityCache::TestCase
```
dylanahsmith: `@record.save` in the setup would cause the cache key to be memoized, so I wanted to use an empty setup at least for test_unsupported_joins_in_assocation_scope.

Thibaut:
:+1:

lib/identity_cache/configuration_dsl.rb
```diff
@@ -291,13 +291,9 @@ def ensure_cacheable_association(association, options)
...
           raise UnsupportedAssociationError, "caching through associations isn't supported"
         end
-        child_class = association_reflection.klass
-        scope = association_reflection.scope
-        if scope && !child_class.all.instance_exec(&scope).joins_values.empty?
-          raise UnsupportedAssociationError, "caching association scoped with a join isn't supported"
-        end
         options[:inverse_name] ||= association_reflection.inverse_of.name if association_reflection.inverse_of
         options[:inverse_name] ||= self.name.underscore.to_sym
+        child_class = association_reflection.klass
```
richardmonette: Was the order of this line changed on purpose?

richardmonette:
:+1:

lib/identity_cache/configuration_dsl.rb
dylanahsmith: It was moved back, so basically a revert of https://github.com/Shopify/identity_cache/pull/259/files#diff-6f09f93a419109ba1acdc9d10e897513 for this file

PR #261, opened 2016-02-26
Avoid re-fetching a cache_belongs_to association in prefetch_associations

dylanahsmith:

## Problem

There was a bug in prefetch_associations (introduced in pull #252) where it didn't check if a cache_belongs_to association was already fetched before fetching it, so the association can end up being double fetched.

## Solution

Check if the instance variable that holds the cached record has been set and avoid calling fetch_multi to get it if it is there.


eapache:
:+1:

boourns:
:+1:

PR #262, opened 2016-02-29
fix(prefetch_associations): Avoid exists query on a collection proxy.

dylanahsmith:

## Problem

When passing an unloaded association (i.e. a ActiveRecord::Associations::CollectionProxy object) directly into `prefetch_associations` as the records parameter, the `records.empty?` check causes an `exists?` query.  This is useless query, since the association is still going to be loaded after that empty check.

## Solution

Convert `records` to an array to avoid the extra query.

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -75,6 +75,7 @@ module IdentityCache
       end
 
       def prefetch_associations(associations, records)
+        records = records.to_a
         return if records.empty?
         unless IdentityCache.should_use_cache?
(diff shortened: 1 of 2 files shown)


eapache:
Unless the association is actually empty, in which case we save the cost of the full loading query by running the (presumably faster) exists query and short-circuiting. However I can't imagine that is a common case, so this LGTM.

dylanahsmith:
If we are preloading the association, then I assume it will be accessed after prefetch_associations is called, so even with the short-circuiting it is still likely to need the full query.  Also, I doubt the exists query is much faster when the full query doesn't match any rows.

eapache:
:+1:

boourns:
:+1:

PR #263, opened 2016-02-29
fix(prefetch_associations): Preload using database in a transaction.

dylanahsmith:

## Problem

If prefetch_associations gets called in a transaction, then it would load the associations from the cache and store them in an instance variable that the cached accessors (i.e. `fetch_#{association}`) wouldn't use inside the transaction.  This will also lead to an N+1 when the code uses it with the assumption that the cached associations are loaded.

## Solution

Use active record to preload the associations when `IdentityCache.should_use_cache?` returns false.

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -76,6 +76,14 @@ module IdentityCache
 
       def prefetch_associations(associations, records)
         return if records.empty?
+        unless IdentityCache.should_use_cache?
+          if ActiveRecord::VERSION::MAJOR == 4 && ActiveRecord::VERSION::MINOR == 0
+            ActiveRecord::Associations::Preloader.new(records, associations).run
+          else
+            ActiveRecord::Associations::Preloader.new.preload(records, associations)
+          end
+          return
+        end
 
         case associations
         when nil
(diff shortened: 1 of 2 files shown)


eapache:
I don't know enough about the interaction of IdentityCache and ActiveRecord to feel super-comfortable blessing this, but the logic makes sense, so tentative :+1:.

Some general questions not really specific to this PR but more for my understanding:
- What causes the instance variable name to change when inside a transaction or not? I don't see obvious code for that, or an obvious reason.
- Why is ActiveRecord version exactly 4.0 (but not any other versions before/after?) so special?

dylanahsmith:
> What causes the instance variable name to change when inside a transaction or not?

It doesn't affect the instance variable, it causes `IdentityCache.should_use_cache?` to return false, which causes `fetch_#{association}` to ignore the instance variable and only use the active record association.

> Why is ActiveRecord version exactly 4.0 (but not any other versions before/after?) so special?

We don't support rails 3.2, so I didn't bother testing it, and wanted to make it clear that the code could be removed once we stop supporting rails 4.0.

eapache:
Makes sense now, more confident :+1:

boourns:
:+1:

PR #265, opened 2016-03-16
Fix warnings

hirocaster:
Fix <del>all</del> warnings.

### DIFF
--- a/lib/identity_cache.rb
+++ b/lib/identity_cache.rb
@@ -58,7 +58,7 @@ module IdentityCache
     # +cache_adaptor+ - A ActiveSupport::Cache::Store
     #
     def cache_backend=(cache_adaptor)
-      if @cache
+      if defined?(@cache)
         cache.cache_backend = cache_adaptor
       else
         @cache = MemoizedCacheProxy.new(cache_adaptor)
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -344,8 +344,8 @@ module IdentityCache
               parent_record_to_child_records[parent_record] << child_record
             end
 
-            parent_record_to_child_records.each do |parent_record, child_records|
-              parent_record.send(details[:prepopulate_method_name], child_records)
+            parent_record_to_child_records.each do |parent, children|
+              parent.send(details[:prepopulate_method_name], children)
             end
           end
 
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -319,7 +319,7 @@ module IdentityCache
         first_record = records.first
         return if first_record.association(association).loaded?
         iv_name_key = details[:embed] == true ? :records_variable_name : :ids_variable_name
-        return if first_record.instance_variable_get(:"@#{details[iv_name_key]}")
+        return if first_record.instance_variable_defined?(:"@#{details[iv_name_key]}")
         fetch_embedded_associations(records)
       end
 
--- a/test/recursive_denormalized_has_many_test.rb
+++ b/test/recursive_denormalized_has_many_test.rb
@@ -61,7 +61,7 @@ class RecursiveDenormalizedHasManyTest < IdentityCache::TestCase
       # one for the mid level has_one association
       # one for the deep level level has_many on the mid level has_many association
       # one for the deep level level has_many on the mid level has_one association
-      record_from_cache_miss = Item.fetch(@record.id)
+      Item.fetch(@record.id)
     end
   end
 
@@ -103,7 +103,7 @@ class RecursiveNormalizedHasManyTest < IdentityCache::TestCase
 
   def test_cache_repopulation_should_not_fetch_non_embedded_associations
     Item.any_instance.expects(:fetch_associated_records).never
-    record_from_cache_miss = Item.fetch(@record.id)
+    Item.fetch(@record.id) # cache miss
   end
 end
 
--- a/test/schema_change_test.rb
+++ b/test/schema_change_test.rb
@@ -103,7 +103,6 @@ class SchemaChangeTest < IdentityCache::TestCase
     Item.cache_has_many :polymorphic_records, :inverse_name => :owner, :embed => :ids
     read_new_schema
 
-    Item.expects(:resolve_cache_miss).returns(@record)
     record = Item.fetch(@record.id)
   end
 end
--- a/test/schema_change_test.rb
+++ b/test/schema_change_test.rb
@@ -89,10 +89,12 @@ class SchemaChangeTest < IdentityCache::TestCase
 
     record = Item.fetch(@record.id)
 
+    teardown_models
+    setup_models
+
     Item.cache_has_many :polymorphic_records, :inverse_name => :owner, :embed => true
     read_new_schema
 
-    Item.expects(:resolve_cache_miss).returns(@record)
     record = Item.fetch(@record.id)
   end
 
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -11,7 +11,7 @@ require 'active_support/cache/memcached_store'
 require File.dirname(__FILE__) + '/../lib/identity_cache'
 
 DatabaseConnection.setup
-ActiveSupport::Cache::Store.instrument = true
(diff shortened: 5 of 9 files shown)


lib/identity_cache.rb
```diff
@@ -58,7 +58,7 @@ def included(base) #:nodoc:
     # +cache_adaptor+ - A ActiveSupport::Cache::Store
     #
     def cache_backend=(cache_adaptor)
-      if @cache
+      if instance_variable_defined?("@cache")
```
dylanahsmith: `if defined?(@cache)` is faster and more succinct

lib/identity_cache/configuration_dsl.rb
```diff
@@ -205,11 +205,20 @@ def build_recursive_association_cache(association, options) #:nodoc:
...
         options[:records_variable_name]  = "cached_#{association}"
 
-        self.class_eval(<<-CODE, __FILE__, __LINE__ + 1)
-          def #{options[:cached_accessor_name]}
-            fetch_recursively_cached_association('#{options[:records_variable_name]}', :#{association})
-          end
-        CODE
+        if method_defined?(options[:cached_accessor_name].to_sym)
+          self.class_eval(<<-CODE, __FILE__, __LINE__ + 1)
+            def #{options[:cached_accessor_name]}_with_idc
```
dylanahsmith: why is this needed?

lib/identity_cache/query_api.rb
```diff
@@ -399,10 +405,12 @@ def fetch_recursively_cached_association(ivar_name, association_name) # :nodoc:
...
         ivar_full_name = :"@#{ivar_name}"
 
-        unless ivar_value = instance_variable_get(ivar_full_name)
-          ivar_value = IdentityCache.map_cached_nil_for(send(association_name))
-          instance_variable_set(ivar_full_name, ivar_value)
-        end
+        ivar_value = if instance_variable_defined?(ivar_full_name)
+                       instance_variable_get(ivar_full_name)
+                     else
+                       value = IdentityCache.map_cached_nil_for(send(association_name))
```
dylanahsmith: `IdentityCache.map_cached_nil_for` was to distinguish between `nil` and unloaded.  I think we should instead use the instance variable defined check to distinguish between these two states.  Since that isn't an isolated change, it should probably be handled in a separate PR.

lib/identity_cache/query_api.rb
```diff
@@ -344,8 +350,8 @@ def prefetch_one_association(association, records)
               parent_record_to_child_records[parent_record] << child_record
             end
 
-            parent_record_to_child_records.each do |parent_record, child_records|
-              parent_record.send(details[:prepopulate_method_name], child_records)
+            parent_record_to_child_records.each do |parent, child|
```
dylanahsmith: I think `children` is more appropriate than `child` since it is an array.

lib/identity_cache/query_api.rb
```diff
@@ -319,7 +324,8 @@ def prefetch_embedded_association(records, association, details)
         first_record = records.first
         return if first_record.association(association).loaded?
         iv_name_key = details[:embed] == true ? :records_variable_name : :ids_variable_name
-        return if first_record.instance_variable_get(:"@#{details[iv_name_key]}")
+        iv_name = :"@#{details[iv_name_key]}"
+        return if first_record.instance_variable_defined?(iv_name) && first_record.instance_variable_get(iv_name)
```
dylanahsmith: I think we only need to check if the instance variable exists

test/recursive_denormalized_has_many_test.rb
```diff
@@ -61,7 +61,7 @@ def test_on_cache_miss_child_record_fetch_should_include_nested_associations_to_
       # one for the mid level has_one association
       # one for the deep level level has_many on the mid level has_many association
       # one for the deep level level has_many on the mid level has_one association
-      record_from_cache_miss = Item.fetch(@record.id)
+      Item.fetch(@record.id) # record_from_cache_miss
```
dylanahsmith: I think `# cache miss` would be a more appropriate comment.  Although I don't think a cache miss comment is necessary since the test name makes it clear that it is testing a cache miss.

test/recursive_denormalized_has_many_test.rb
```diff
@@ -103,7 +103,7 @@ def setup
 
   def test_cache_repopulation_should_not_fetch_non_embedded_associations
     Item.any_instance.expects(:fetch_associated_records).never
-    record_from_cache_miss = Item.fetch(@record.id)
+    Item.fetch(@record.id) # record_from_cache_miss
```
dylanahsmith: I think `# cache miss` would be appropriate here

test/test_helper.rb
```diff
@@ -11,7 +11,6 @@
 require File.dirname(__FILE__) + '/../lib/identity_cache'
 
 DatabaseConnection.setup
-ActiveSupport::Cache::Store.instrument = true
```
dylanahsmith: I think this is causing test failures.  E.g. `0 instead of 1 memcache operations were executed` I think is because it needs this instrumentation to count memcache operations

lib/identity_cache/configuration_dsl.rb
hirocaster: this warning is `warning: method redefined; discarding old fetch_polymorphic_records`

Exists `#fetch_polymorphic_records` by AR.
`self.class_eval` is `#fetch_polymorphic_records` redefined.

This approach 

Define method(suffix `with_idc`), if exists method.
It using by alias_method_chain.

lib/identity_cache/query_api.rb
hirocaster: OK, these changes remove in PR.
I will track #266.

test/recursive_denormalized_has_many_test.rb
hirocaster: I think so too. Therefore remove cache miss comment.

hirocaster:
Thank you code review, good speedy feedback. 
Fixed, please confirm.

lib/identity_cache/configuration_dsl.rb
dylanahsmith: Active record doesn't define fetch_polymorphic_records.  It looks like this is coming from SchemaChangeTest#test_embed_existing_cache_has_many which is making sure it is safe to change an association between being fully embedded or just embedding the id.  This warning is an indication of a problem in the tests, not the code under test.

Perhaps that can be done more cleanly using

```ruby
teardown_models
setup_models
```

which is done between tests.

lib/identity_cache/query_api.rb
```diff
@@ -182,7 +182,12 @@ def add_cached_associations_to_coder(record, coder)
           end
           if (cached_has_manys = klass.cached_has_manys).present?
             coder[:association_ids] = cached_has_manys.each_with_object({}) do |(name, options), hash|
-              hash[name] = record.instance_variable_get(:"@#{options[:ids_variable_name]}") unless options[:embed] == true
+              unless options[:embed] == true
+                iv_name = :"@#{options[:ids_variable_name]}"
+                hash[name] = if record.instance_variable_defined?(iv_name)
```
dylanahsmith: It looks like this check is unnecessary, and was just added to workaround a warning in another specific test. SchemaChangeTest#test_add_non_embedded_cache_has_many does `Item.expects(:resolve_cache_miss).returns(@record)` which mocks out the code that sets this variable.

lib/identity_cache/configuration_dsl.rb
hirocaster: Thank you, comment. :smile: 

I understand define fetch_polymorphic_records. 
I red

- ActiveRecordObjects
- #build_recursive_association_cache
- #build_id_embedded_has_many_cache

:eyes: 
--

Thats is certainly redefin #fetch_polymorphic_records in SchemaChangeTest#test_embed_existing_cache_has_many.

test/schema_change_test.rb
```diff
@@ -89,10 +89,12 @@ def test_embed_existing_cache_has_many
 
     record = Item.fetch(@record.id)
 
+    teardown_models
+    setup_models
```
hirocaster: I think not best. :disappointed:

dylanahsmith:
Thanks for the fixes

Issue #163, opened 2014-06-11
Remove :idc_cached_nil


fbogsany:
@dylanahsmith : "I think the only thing that prevented us from storing `nil` was that `cache.read` would return `nil` for a cache miss as well as if the value cached was `nil`. I think we may even be able to get rid of `map_cached_nil_for` now that we don't use `cache.read`."

Follow-up from #154 

Note that support for `cache.read` was re-added in #161

PR #266, opened 2016-03-16
Use instance_variable_defined? instead of storing mapped nils

dylanahsmith:
For issue #163 and pull #265
cc @hirocaster

@rafaelfranca please review

We needed to distinguish between `nil` and uncached in the response to memcache queries by storing a special symbol instead of `nil` in memcached.  However, we don't need to do the same thing when storing cached values in instance variables, since an undefined instance variable can represent an unloaded value.

This should also help reduce warnings when running with `ruby -w` that `rake` now uses by default, since we won't be accessing these instance variables before they are defined.


rafaelfranca:
I agree this is better behavior :shipit:

Issue #246, opened 2015-11-14
Memoization can cause multiple records to share an attributes hash


eugeneius:
When memoization is enabled, the same coder hash is used to instantiate multiple objects when a record is fetched multiple times. The data is passed to `#init_with` [without being duplicated](https://github.com/Shopify/identity_cache/blob/v0.2.5/lib/identity_cache/query_api.rb#L93), which also just [assigns the hash directly](https://github.com/rails/rails/blob/v4.2.5/activerecord/lib/active_record/core.rb#L303).

Here's a test case demonstrating how this behaviour caused a bug in our application: https://gist.github.com/e644d420da97dfc0b920

The `title` attribute is assigned on one record, but not saved. This updates the value in the attributes hash, but doesn't write it to the database. When the record is fetched again, it gets the same attributes hash, including the updated `title` attribute. We try to update the attribute on this new record, but the new value is the same as the existing one and Active Record optimises away the update.

This problem was actually fixed by #242, as the attributes hash is now duplicated every time a record is created. However the bug is still present in the latest released version, v0.2.5.

Since the worst case here is pretty bad (lost updates!) it would be great to either publish a new release from master to get the fix in #242, or do a bugfix release just to address this issue.

camilo:
Sounds like #242 is on the latest release now.

eugeneius:
:+1: thanks for following up, @camilo.

Do you think it's worthwhile adding the test case I wrote above to prevent this from regressing?

camilo:
@eugeneius I think so yes, if you can turn it into a PR that'd be even better :)

eugeneius:
It turns out this wasn't totally fixed by https://github.com/Shopify/identity_cache/pull/242 :grimacing: 

I've opened https://github.com/Shopify/identity_cache/pull/267 to fix the remaining case where this bug can happen.

PR #267, opened 2016-04-03
Copy attributes hash before writing memoized cache

eugeneius:
These tests will fail if [the code that copies the attributes hash before instantiating the model](https://github.com/Shopify/identity_cache/blob/v0.3.1/lib/identity_cache/query_api.rb#L88) is removed.

Removing that code would cause multiple records to share an attributes hash when memoization is enabled, which would break dirty tracking. This was the case in versions before v0.3.0: https://github.com/Shopify/identity_cache/issues/246

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # IdentityCache changelog
 
+#### Unreleased
+
+- Stop sharing the same attributes hash between the fetched record and the memoized cache, which could interfere with dirty tracking
+
 #### 0.3.1
 
 - Fix cache_index for non-id primary key
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -164,7 +164,7 @@ module IdentityCache
       def coder_from_record(record) #:nodoc:
         unless record.nil?
           coder = {
-            attributes: record.attributes_before_type_cast,
+            attributes: record.attributes_before_type_cast.dup,
             class: record.class,
           }
           add_cached_associations_to_coder(record, coder)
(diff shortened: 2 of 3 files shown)


eugeneius:
It looks like this is actually still broken. ¯\\\_(ツ)_/¯

I'll try to figure out what's going on.

eugeneius:
Okay, I understand what's happening now.

When I originally wrote the test to demonstrate https://github.com/Shopify/identity_cache/issues/246, it looked like this:

```ruby
Item.fetch(item.id)
Item.fetch(item.id).title = "my title"
Item.fetch(item.id).update(title: "my title")
```

The first call to `fetch` is filling the cache so that the records returned from the other two calls share the same attributes hash and demonstrate the bug. As of https://github.com/Shopify/identity_cache/pull/242 that bug is fixed, as the attributes hash is now always copied before being used to instantiate a record.

When I was opening this pull request to add my test to catch regressions, several months had passed and I had forgotten the exact details of the problem. I removed the first `fetch` since it seemed unnecessary. This accidentally exposed a related but different bug: when the cache doesn't have an entry for a record, we fetch it from the database and write the result back to the cache. With memoization enabled, this would result in the attributes hash in the cache being shared with the record returned to the application. Updating the record would also change the cache entry, and fetching the record again could result in it being instantiated with dirty attributes that don't accurately reflect what's in the database.

The fix is similar to the previous one: copy the attributes hash before writing it back to the cache.

camilo:
thanks, this LGTM, @dylanahsmith / @rafaelfranca want :eyes: would be appreciated as well.

rafaelfranca:
:+1:

camilo:
@eugeneius mind updating the CHANGELOG ? I'll merge after, thanks!

eugeneius:
Done!

PR #271, opened 2016-04-26
Fix id mismatch check for string primary keys.

### LINKED ISSUE #270: Non-integer primary keys
If model uses non-integer primary key (UUID primary key for example) there is an id mismatch log entry like this:

```
[IdentityCache] (cache_backend) cache hit for IDC:6:blob:CustomerSession:16050426020716856423:caa8f134-0904-41dd-842d-8a574466a6a4
[IDC id mismatch] fetch_by_id_requested=caa8f134-0904-41dd-842d-8a574466a6a4 fetch_by_id_got=caa8f134-0904-41dd-842d-8a574466a6a4 
```

dylanahsmith:
@camilo 

Fixes #270

Coerce integers to strings rather than coercing string to integers for the mismatch check.


camilo:
👍 a regression test would be nice, but not 100% needed in this case

PR #276, opened 2016-06-14
Configurable fetch association behaviour [return array/relation]

daniellaniyo:
Fetch association to return array or relation

@dylanahsmith & @tjoyal for review/thoughts. Cc @camilo 

## Problem

There is an [open PR](https://github.com/Shopify/identity_cache/pull/224) that prevents chaining by having fetch association to return an array. However, it was never merged since it requires changes on the Shopify side. The idea of this PR to have a version where fetch association returns an array but also supports the current behaviour, i.e. returning a relation. This will allow us to replace these relations dependencies gradually on the Shopify side while having the correct IDC default behaviour.

Once everything is cleaned up, we will plan a major release that only returns arrays.

## Solution

By default, return an Array from the fetch association method rather than a relation to prevent chaining.  Any time we need to support the old behaviour, use `IdentityCache.allow_fetch_to_return_relation` around the block. 

A first test was done on [this shopify branch](https://github.com/Shopify/shopify/compare/bump-IDC) to confirm that [tests pass](https://buildkite.com/shopify/shopify-branches/builds/112661) with these changes.  `IdentityCache.allow_fetch_to_return_relation`  will be used around tests that still depend on relations returned by fetch_association.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 # IdentityCache changelog
 
-#### Unreleased
+#### 0.3.2
 
-- Stop sharing the same attributes hash between the fetched record and the memoized cache, which could interfere with dirty tracking
+- Fetch association returns relation or array depending on the configuration. It was only returning a relation for cache_has_many fetch association methods. (#276)
+- Stop sharing the same attributes hash between the fetched record and the memoized cache, which could interfere with dirty tracking (#267)
 
 #### 0.3.1
 
--- a/lib/identity_cache.rb
+++ b/lib/identity_cache.rb
@@ -40,6 +40,9 @@ module IdentityCache
     mattr_accessor :cache_namespace
     self.cache_namespace = "IDC:#{CACHE_VERSION}:".freeze
 
+    mattr_accessor :fetch_returns_relation
+    self.fetch_returns_relation = true
+
     def included(base) #:nodoc:
       raise AlreadyIncludedError if base.include?(IdentityCache::ConfigurationDSL)
 
@@ -129,6 +132,14 @@ module IdentityCache
       end
 
       result
+    end
+
+    def with_fetch_returns_relation(value = true)
+      previous_value = self.fetch_returns_relation
+      self.fetch_returns_relation = value
+      yield
+    ensure
+      self.fetch_returns_relation = previous_value
     end
 
     private
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -236,7 +236,11 @@ module IdentityCache
             if IdentityCache.should_use_cache? || #{association}.loaded?
               @#{options[:records_variable_name]} ||= #{options[:association_reflection].klass}.fetch_multi(#{options[:cached_ids_name]})
             else
-              #{association}
+              if IdentityCache.fetch_returns_relation
+                #{association}
+              else
+                #{association}.to_a
+              end
             end
           end
 
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -156,7 +156,7 @@ module IdentityCache
 
       def get_embedded_association(record, association, options) #:nodoc:
         embedded_variable = record.public_send(options.fetch(:cached_accessor_name))
-        if record.class.reflect_on_association(association).collection?
+        if embedded_variable.respond_to?(:to_ary)
           embedded_variable.map {|e| coder_from_record(e) }
         else
           coder_from_record(embedded_variable)
@@ -398,7 +398,7 @@ module IdentityCache
     private
 
     def fetch_recursively_cached_association(ivar_name, association_name) # :nodoc:
-      if IdentityCache.should_use_cache?
+      assoc = if IdentityCache.should_use_cache?
         ivar_full_name = :"@#{ivar_name}"
 
         if instance_variable_defined?(ivar_full_name)
@@ -409,6 +409,8 @@ module IdentityCache
       else
         send(association_name.to_sym)
       end
+      assoc = assoc.to_ary if assoc.respond_to?(:to_ary) && !IdentityCache.fetch_returns_relation
+      assoc
     end
 
     def expire_primary_index # :nodoc:
--- a/test/denormalized_has_many_test.rb
+++ b/test/denormalized_has_many_test.rb
@@ -12,6 +12,14 @@ class DenormalizedHasManyTest < IdentityCache::TestCase
     @record.reload
   end
 
+  def test_uncached_record_from_the_db_should_come_back_with_association_array_when_fetch_returns_array
+    IdentityCache.with_fetch_returns_relation(false) do    
+      assert_equal IdentityCache.fetch_returns_relation, false
+      record_from_db = Item.find(@record.id)
+      assert_equal Array, record_from_db.fetch_associated_records.class
+    end
+  end
+
   def test_uncached_record_from_the_db_will_use_normal_association
     expected = @record.associated_records
     record_from_db = Item.find(@record.id)
@@ -20,6 +28,17 @@ class DenormalizedHasManyTest < IdentityCache::TestCase
 
     assert_equal @record, record_from_db
     assert_equal expected, record_from_db.fetch_associated_records
+  end
+
+  def test_on_cache_hit_record_should_come_back_with_cached_association_array_when_fetch_returns_array
+    IdentityCache.with_fetch_returns_relation(false) do  
+      assert_equal IdentityCache.fetch_returns_relation, false
+      Item.fetch(@record.id) # warm cache
+
+      record_from_cache_hit = Item.fetch(@record.id)
+      assert_equal @record, record_from_cache_hit
+      assert_equal Array, record_from_cache_hit.fetch_associated_records.class
+    end
   end
 
   def test_on_cache_hit_record_should_come_back_with_cached_association
@@ -93,6 +112,14 @@ class DenormalizedHasManyTest < IdentityCache::TestCase
     IdentityCache.cache.expects(:delete).with(@record.primary_cache_index_key)
     child.save!
   end
+
+  def test_fetch_association_does_not_allow_chaining_when_fetch_returns_array
+    IdentityCache.with_fetch_returns_relation(false) do
+      check = proc { assert_equal false, Item.fetch(@record.id).fetch_associated_records.respond_to?(:where) }
+      2.times { check.call } # for miss and hit
+      Item.transaction { check.call }
+    end
+  end  
 
   class CheckAssociationTest < IdentityCache::TestCase
     def test_unsupported_through_assocation
--- a/lib/identity_cache/version.rb
+++ b/lib/identity_cache/version.rb
@@ -1,4 +1,4 @@
 module IdentityCache
-  VERSION = "0.3.1"
+  VERSION = "0.3.2"
   CACHE_VERSION = 6
 end
--- a/test/normalized_has_many_test.rb
+++ b/test/normalized_has_many_test.rb
@@ -195,4 +195,11 @@ class NormalizedHasManyTest < IdentityCache::TestCase
     @not_cached.save!
   end
 
+  def test_fetch_association_does_not_allow_chaining_when_fetch_returns_array
+    IdentityCache.with_fetch_returns_relation(false) do
+      check = proc { assert_equal false, Item.fetch(@record.id).fetch_associated_records.respond_to?(:where) }
+      2.times { check.call } # for miss and hit
+      Item.transaction { check.call }
+    end
+  end
 end


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 - Stop sharing the same attributes hash between the fetched record and the memoized cache, which could interfere with dirty tracking
```
dylanahsmith: wouldn't this be part of the release too?

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
...
 
+#### 0.3.2
+
+- Fetch association returns relation or array depending on the configuration
```
dylanahsmith: This isn't clear about how it is configured by default.  It should behave the same way by default until 0.4 is released.

lib/identity_cache.rb
```diff
@@ -40,6 +40,9 @@ class << self
     mattr_accessor :cache_namespace
     self.cache_namespace = "IDC:#{CACHE_VERSION}:".freeze
 
+    mattr_accessor :fetch_returns_relation
+    self.fetch_returns_relation = false
```
dylanahsmith: This should be true to avoid having this release be a breaking change

lib/identity_cache/configuration_dsl.rb
```diff
@@ -236,7 +236,11 @@ def #{options[:cached_accessor_name]}
             if IdentityCache.should_use_cache? || #{association}.loaded?
               @#{options[:records_variable_name]} ||= #{options[:association_reflection].klass}.fetch_multi(#{options[:cached_ids_name]})
             else
-              #{association}
+              if IdentityCache.fetch_returns_relation
+                #{association}
+              else
+                association(:#{association}).load_target
```
dylanahsmith: I think `#{association}.to_a` would be preferred since that is part of the documented active record API.

lib/identity_cache/query_api.rb
```diff
@@ -398,18 +404,30 @@ def prefetch_one_association(association, records)
...
-
-        assoc.is_a?(ActiveRecord::Associations::CollectionAssociation) ? assoc.reader : assoc
       else
-        send(association_name.to_sym)
+        if IdentityCache.should_use_cache?
+          ivar_full_name = :"@#{ivar_name}"
+          if instance_variable_defined?(ivar_full_name)
+            instance_variable_get(ivar_full_name)
+          else
+            instance_variable_set(ivar_full_name, association(association_name).load_target)
```
dylanahsmith: I think it will be easier to reason about the code if we always store the samr thing in this instance variable and just call to_a on the result if IdentityCache.fetch_returns_relation returns false

CHANGELOG.md
dylanahsmith: We might also want to mention that it was only returning a relation for cache_has_many fetch association methods

lib/identity_cache/query_api.rb
```diff
@@ -398,18 +404,30 @@ def prefetch_one_association(association, records)
...
+          else
+            instance_variable_set(ivar_full_name, send(association_name))
+          end
 
-        assoc = if instance_variable_defined?(ivar_full_name)
-          instance_variable_get(ivar_full_name)
+          assoc.is_a?(ActiveRecord::Associations::CollectionAssociation) ? assoc.reader : assoc
         else
-          instance_variable_set(ivar_full_name, send(association_name))
+          send(association_name.to_sym)
```
camilo: `public_send` ?

daniellaniyo:
Thanks for your feedback! Please check [latest commit](https://github.com/Shopify/identity_cache/pull/276/commits/0a16f2cb38bdda9e9bfae45b9cebd01a30fc10be?diff=split) for changes.

CHANGELOG.md
daniellaniyo: Yes, it will! I have updated the changelog in my last commit.

lib/identity_cache.rb
daniellaniyo: Done! Now, set to false. I have added an `allow_fetch_to_return_array` method to go along with this change.

camilo:
LGTM wait for @dylanahsmith tho

lib/identity_cache.rb
```diff
@@ -131,6 +134,18 @@ def fetch_multi(*keys)
       result
     end
 
+    def allow_fetch_to_return_relation
+      self.fetch_returns_relation = true
+      yield
+      self.fetch_returns_relation = false
```
dylanahsmith: To avoid leaky behaviour, we usually store the old value in a local variable then use an ensure block to set it back to that old value.

You could also merge this function with the following one by just parameterizing the value.  E.g. `def with_ fetch_returns_relation(value=true)`

lib/identity_cache/query_api.rb
```diff
@@ -156,7 +156,13 @@ def set_embedded_association(record, association_name, coder_or_array) #:nodoc:
...
       def get_embedded_association(record, association, options) #:nodoc:
         embedded_variable = record.public_send(options.fetch(:cached_accessor_name))
-        if record.class.reflect_on_association(association).collection?
+        cached_association = if IdentityCache.fetch_returns_relation 
+          record.class.reflect_on_association(association).collection?
+        else
+          embedded_variable.is_a?(Array)
+        end
+
+        if cached_association
```
dylanahsmith: What you really care about is if embedded_variable is array like, which can be checked for with just `if embedded_variable.respond_to?(:to_ary)`, then you can use `embedded_variable.to_ary.map {`... on the following line

lib/identity_cache/query_api.rb
```diff
@@ -400,16 +406,18 @@ def prefetch_one_association(association, records)
...
       if IdentityCache.should_use_cache?
         ivar_full_name = :"@#{ivar_name}"
-
         assoc = if instance_variable_defined?(ivar_full_name)
           instance_variable_get(ivar_full_name)
         else
           instance_variable_set(ivar_full_name, send(association_name))
         end
-
-        assoc.is_a?(ActiveRecord::Associations::CollectionAssociation) ? assoc.reader : assoc
```
dylanahsmith: This is a weird case.  It doesn't look like `send(association_name)` would return a ActiveRecord::Associations::CollectionAssociation.  I think this is working around an inconsistency that shouldn't be there.  I'm going to open up a pull request to fix this, since it looks like it is causing some confusion.

lib/identity_cache/query_api.rb
```diff
@@ -400,16 +406,18 @@ def prefetch_one_association(association, records)
...
+        if IdentityCache.fetch_returns_relation
+          assoc.is_a?(ActiveRecord::Associations::CollectionAssociation) ? assoc.reader : assoc
+        else
+          assoc.to_a
```
dylanahsmith: This method is used for cache_has_one's fetch association method as well, so we can't always call `.to_a` if `IdentityCache.fetch_returns_relation` is false

lib/identity_cache/query_api.rb
dylanahsmith: See https://github.com/Shopify/identity_cache/pull/277

test/denormalized_has_many_test.rb
```diff
@@ -12,7 +12,23 @@ def setup
     @record.reload
   end
 
+  def test_uncached_record_from_the_db_will_use_normal_association_as_ary
+    # Test configuration where fetch association returns an array
```
dylanahsmith: Try to make the code more readable by using more appropriate variable names before adding comments.  For instance, the test name doesn't mention the configuration or `allow_fetch_to_return_array`, but it looks like that is what you are actually testing.

test/denormalized_has_many_test.rb
```diff
@@ -12,7 +12,23 @@ def setup
...
+    IdentityCache.allow_fetch_to_return_array do    
+      assert_equal IdentityCache.fetch_returns_relation, false
+      expected = @record.associated_records.to_a
+      record_from_db = Item.find(@record.id)
+      assert_queries(1) do
+        assert_memcache_operations(0) do
```
dylanahsmith: I think you are trying to test too many things here.  Try to keep the test cases more focused.  I don't think you need to test the number of queries in this case.

test/denormalized_has_many_test.rb
```diff
@@ -12,7 +12,23 @@ def setup
...
+          assert_equal expected, record_from_db.fetch_associated_records
```
dylanahsmith: It looks like a collection proxy implements `==` to compare the loaded target to the other object, so I don't think this will actually catch a regression.  I think the assertion you actually want is `assert_equal Array, @record.fetch_associated_records.class`

test/denormalized_has_many_test.rb
```diff
@@ -12,7 +12,23 @@ def setup
...
+        end
+      end
+    end
+  end
+
   def test_uncached_record_from_the_db_will_use_normal_association
+    # Test configuration where fetch association returns a relation
+    assert_equal IdentityCache.fetch_returns_relation, true
```
dylanahsmith: I don't think this is relevant for this test.

test/denormalized_has_many_test.rb
dylanahsmith: Actually, it looks like the denormalized and normalized test_fetch_association_does_not_allow_chaining_with_ary tests already provide regression tests.  What else are you trying to test with these other tests?

lib/identity_cache/query_api.rb
daniellaniyo: Would using `load_target` be a good idea here for cases where `IdentityCache.fetch_returns_relation` is false?

dylanahsmith: No, I think that should be handled separately.  I think this is what we need here

```ruby
assoc = assoc.reader if assoc.is_a?(ActiveRecord::Associations::CollectionAssociation)
assoc = assoc.to_ary if assoc.respond_to?(:to_ary) && !IdentityCache.fetch_returns_relation
assoc
```

test/denormalized_has_many_test.rb
daniellaniyo: I could surely focus on the class of the fetch result.

I had the number of queries checked to make sure we had the right behaviour for both cached & uncached records; and ensure tests in the previous [PR](https://github.com/Shopify/identity_cache/pull/224/files) were still passing.

lib/identity_cache/query_api.rb
dylanahsmith: Actually, the `assoc = assoc.to_ary if assoc.respond_to?(:to_ary) && !IdentityCache.fetch_returns_relation` line should probably happen outside of `if IdentityCache.should_use_cache?`

daniellaniyo:
@dylanahsmith Could you please check out [the latest commit](https://github.com/Shopify/identity_cache/pull/276/commits/b09c5f6798189f641b823d890eca966cc7609d4a)? Thoughts?

daniellaniyo:
Rebased to grab [277](https://github.com/Shopify/identity_cache/pull/277) changes and squashed previous commits. @dylanahsmith Could you please check out [the latest commit](https://github.com/Shopify/identity_cache/pull/276/files?diff=split)? Thoughts?

lib/identity_cache.rb
```diff
@@ -131,6 +134,14 @@ def fetch_multi(*keys)
       result
     end
 
+    def with_fetch_returns_relation(value = true)
+      previous_value = self.fetch_returns_relation
+      self.fetch_returns_relation = value
+      yield
+      ensure
```
dylanahsmith: This should be indented only as much as `def`.  The following line also needs less indentation

lib/identity_cache/query_api.rb
```diff
@@ -401,13 +401,15 @@ def fetch_recursively_cached_association(ivar_name, association_name) # :nodoc:
       if IdentityCache.should_use_cache?
```
dylanahsmith: move `assoc = ` to this line and 

```ruby
        assoc = assoc.to_ary if assoc.respond_to?(:to_ary) && !IdentityCache.fetch_returns_relation
        assoc
```

after this condition.  That way the else block can be changed back to `send(association_name.to_sym)` and the conditional array coercion only needs to happen in one place in this method

CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 # IdentityCache changelog
 
-#### Unreleased
+#### 0.3.2
 
+- Fetch association returns relation or array depending on the configuration. It was only returning a relation for cache_has_many fetch associtions methods.
```
dylanahsmith: typo: `associtions` -> `association`

Also, I think we should start referencing pull requests or issues in changelog entries by adding `(#276)` to the end to entry

daniellaniyo: Done. I've added the PR numbers for the 0.3.2 entries. See commit below.

lib/identity_cache.rb
daniellaniyo: Aaaahh my bad...derp

daniellaniyo:
@dylanahsmith Thanks for the review! Please see the latest commit [here](https://github.com/Shopify/identity_cache/pull/276/commits/8aeb2e9307edb8a690ce620efb5b75515914ec70).

dylanahsmith:
:+1:

daniellaniyo:
Great! Squashing and merging.

PR #277, opened 2016-06-14
Consistently store the collection proxy for has many associations

dylanahsmith:
@daniellaniyo & @tjoyal for review, cc @camilo 

## Problem

IdentityCache::QueryApi#get_embedded_association was storing an active record collection proxy in the cached association instance variable, but IdentityCache::QueryApi#set_embedded_association was storing the association itself.  This resulted in an otherwise unnecessary `.is_a?(ActiveRecord::Associations::CollectionAssociation)` when using that instance variable.

## Solution

Store `association.reader` in set_embedded_association so the collection proxy is always stored for cache has many associations

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -146,7 +146,7 @@ module IdentityCache
           association = reflection.association_class.new(record, reflection)
           association.target = coder_or_array.map {|e| record_from_coder(e) }
           association.target.each {|e| association.set_inverse_instance(e) }
-          association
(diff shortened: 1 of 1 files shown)


camilo:
Is this going to need transitional code (or a full cache bust) when it ships on a live system?

dylanahsmith:
no, this is just about what is stored in instance variables

daniellaniyo:
LGTM and less confusing indeed.

tjoyal:
:+1:

Issue #274, opened 2016-06-08
Mark IDC fetched records as read_only


dylanahsmith:
cc @airhorns, @camilo

This was attempted before in https://github.com/Shopify/identity_cache/pull/85, which gives a good explanation for the problem:

> IdentityCache is a forever inconsistent datastore. It is a duplication of state across two different storage backends, and unless we somehow confirmed commit in both places before acknowledging, we can't guarantee that the two are in sync. The ruby process could die before it has a chance to issue the memcache DEL, the request could get lost if the connection is interrupted or any number of other things could prevent IDC from reflecting whats truly in the database.
> 
> This being the case, it is sketchy to write something you find in IDC back to the database, because you might be writing out of date data to the database. This is already the case with a MySQL insert in Rails cause you had to have read that data a finite amount of time ago, but that window is usually much smaller, and we have locking primitives and transactions to manage this. With IDC, we don't have that, and we don't have any way of knowing if the data is out of date without asking the database, which entirely defeats the purpose of IDC.
> 
> So, I suggest we return only readonly records from IDC. If you are gonna write a record, first fetch its values from the DB.

One thing I think we should do differently from the pull request, is that records returned from IDC methods when `IdentityCache.should_cache?` returns false should not be marked as read-only.  The problem with marking those records as read-only is that it prevents us from sharing code to use for reading and writing, e.g. calculating some aggregation for a confirmation page which should be consistent with the calculation for the operation itself.

On the other hand, records fetched from the database on a cache miss when `IdentityCache.should_cache?` returns true should be marked as read-only.  This way code that does the wrong thing will fail deterministically, rather than passing on a cache miss and failing on a cache hit.

We might also want to consider making this behaviour configurable, so that we don't get blocked on Shopify needed to be updated to work with this new behaviour, which was probably the main reason that the previous PR didn't get finished and merged.  That way we can at least have identity cache do the right thing by default on the next release with breaking changes.

airhorns:
All excellent points, I agree! It's tempting to let records from the DB miss path be writable since they are technically up to date but I think determinism and sanity is sooo much more important. Good call.

camilo:
I think @daniellaniyo can do this after she finishes with the unchain-ability of resutls

eugeneius:
> One thing I think we should do differently from the pull request, is that records returned from IDC methods when `IdentityCache.should_cache?` returns false should not be marked as read-only.

[...]

PR #279, opened 2016-07-07
Deprecate setting the inverse active record association on cache hit.

dylanahsmith:
@daniellaniyo & @camilo for review

## Problem

I noticed that we were doing `association.target.each {|e| association.set_inverse_instance(e) }` on a cache hit, which meant we were setting the active record association rather than just the cached association.

This is inconsistent with our philosophy of requiring code to explicitly use identity cache, since code like `associated_record.item` can implicitly return a cached record rather than requiring `associated_record.fetch_item`.  This could result in cache corruption accidentally being persisted to the database if the record retrieved through the belongs_to association is saved.

I'm not sure how often this is actually a problem in practice.  That will probably be easier to see once https://github.com/Shopify/identity_cache/issues/274 is implemented.

## Solution

Fixing this would be a breaking change, in fact it will cause test failures in Shopify, so I added IdentityCache.never_set_inverse_association which can be set to true to get the correct behaviour.  In that case it will still set the inverse cached association so `associated_record.item` can be changed to `associated_record.fetch_item` to avoid extra database queries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 #### 0.3.2
 
-- Deprecate setting the inverse active record association on cache hits. Set IdentityCache.never_set_inverse_association to true to avoid this.
+- Deprecate setting the inverse active record association on cache hits. Set IdentityCache.never_set_inverse_association to true to avoid this. (#279)
 - Fetch association returns relation or array depending on the configuration. It was only returning a relation for cache_has_many fetch association methods. (#276)
 - Stop sharing the same attributes hash between the fetched record and the memoized cache, which could interfere with dirty tracking (#267)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 #### 0.3.2
 
+- Deprecate setting the inverse active record association on cache hits. Set IdentityCache.never_set_inverse_association to true to avoid this.
 - Fetch association returns relation or array depending on the configuration. It was only returning a relation for cache_has_many fetch association methods. (#276)
 - Stop sharing the same attributes hash between the fetched record and the memoized cache, which could interfere with dirty tracking (#267)
 
--- a/lib/identity_cache.rb
+++ b/lib/identity_cache.rb
@@ -45,7 +45,7 @@ module IdentityCache
     # fetch_#{association} for a cache_has_many association returns a relation
     # when fetch_returns_relation is set to true and an array when set to false.
     mattr_accessor :fetch_returns_relation
-    self.fetch_returns_relation = version >= Gem::Version.new("0.4")
+    self.fetch_returns_relation = version < Gem::Version.new("0.4")
 
     # Inverse active record associations are set when loading embedded
     # cache_has_many associations from the cache when never_set_inverse_association
--- a/lib/identity_cache.rb
+++ b/lib/identity_cache.rb
@@ -40,8 +40,18 @@ module IdentityCache
     mattr_accessor :cache_namespace
     self.cache_namespace = "IDC:#{CACHE_VERSION}:".freeze
 
+    version = Gem::Version.new(IdentityCache::VERSION)
+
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 #### 0.3.2
 
+- Deprecate setting the inverse active record association on cache hits. Set IdentityCache.never_set_inverse_association to true to avoid this.
```
daniellaniyo: I think the PR number (#279) should be added at the end to keep the logs consistent with the last two entries.

lib/identity_cache.rb
```diff
@@ -40,8 +40,18 @@ class << self
     mattr_accessor :cache_namespace
     self.cache_namespace = "IDC:#{CACHE_VERSION}:".freeze
 
+    version = Gem::Version.new(IdentityCache::VERSION)
+
+    # fetch_#{association} for a cache_has_many association returns a relation
+    # when fetch_returns_relation is set to true and an array when set to false.
     mattr_accessor :fetch_returns_relation
-    self.fetch_returns_relation = true
+    self.fetch_returns_relation = version >= Gem::Version.new("0.4")
```
daniellaniyo: self.fetch_returns_relation should return true for versions < 0.4 to keep it consistent with what we had before ... maybe try `self.fetch_returns_relation = version < Gem::Version.new("0.4")` ?

CHANGELOG.md
dylanahsmith: Thanks for the reminder.  I didn't add it while writing it because I didn't have a PR to get the number from.

lib/identity_cache/query_api.rb
```diff
@@ -145,7 +145,20 @@ def set_embedded_association(record, association_name, coder_or_array) #:nodoc:
...
           association = reflection.association_class.new(record, reflection)
           association.target = coder_or_array.map {|e| record_from_coder(e) }
-          association.target.each {|e| association.set_inverse_instance(e) }
+
+          if reflection.klass < IdentityCache
+            inverse_association_name = record.class.cached_has_manys.fetch(association_name).fetch(:inverse_name)
+            cached_inverse_association = reflection.klass.cached_belongs_tos[inverse_association_name]
+            if cached_inverse_association
+              prepopulate_method_name = cached_inverse_association.fetch(:prepopulate_method_name)
+              association.target.each { |child_record| child_record.send(prepopulate_method_name, record) }
```
daniellaniyo: Does this only set the inverse of the cached association?

dylanahsmith: Yup.  Maybe I should extract this into a private method for clarity.

dylanahsmith:
@daniellaniyo I think I've addressed your feedback now

lib/identity_cache/query_api.rb
```diff
@@ -139,13 +139,31 @@ def record_from_coder(coder) #:nodoc:
...
+      def set_inverse_of_cached_has_many(record, association_reflection, child_records)
+        associated_class = association_reflection.klass
+        return unless associated_class < IdentityCache
+
+        inverse_name = record.class.cached_has_manys.fetch(association_reflection.name).fetch(:inverse_name)
+        inverse_cached_association = associated_class.cached_belongs_tos[inverse_name]
+        return unless inverse_cached_association
+
+        prepopulate_method_name = inverse_cached_association.fetch(:prepopulate_method_name)
+        child_records.each { |child_record| child_record.send(prepopulate_method_name, record) }
```
camilo: is the method ever non-public? if not maybe use `public_send` ?

camilo:
LGMT.

lib/identity_cache/query_api.rb
dylanahsmith: Oh, it looks like we define these methods as public.  However, I don't think they should be part of the public API.  They also do the same thing as `record.instance_variable_set(:"@#{options.fetch(:records_variable_name)}", cached_association)` which we do elsewhere.

PR #280, opened 2016-07-16
Use loaded association first when fetching a cache_has_many association

dylanahsmith:
@daniellaniyo & @csfrancis for review

## Problem

`fetch_#{association}` on a cache_has_many `embed: :ids` association would use the cached association even if the active record association was already loaded.  This can result in an unnecessary query if the cached association wasn't already loaded.  Also, if the active record association was modified, then the stale cached association would get returned.

Lastly, it the cached association would be used if the active record association was loaded and `IdentityCache.should_use_cache?` returns false, which is definitely not desirable.

## Solution

Fixed the condition so that the cached association is only used when using the cache *and* the association is *not* loaded.

There was a test for this intended behaviour, but it was expecting fetch_multi to not be called on the wrong model.  Using `assert_memcache_operations(0)` will catch regressions in a non-fragile way.

### DIFF
--- a/test/cache_invalidation_test.rb
+++ b/test/cache_invalidation_test.rb
@@ -10,6 +10,7 @@ class CacheInvalidationTest < IdentityCache::TestCase
     @record.save
     @record.reload
     @baz, @bar = @record.associated_records[0], @record.associated_records[1]
+    @record.reload
   end
 
   def test_reload_invalidate_cached_ids
--- a/test/cache_invalidation_test.rb
+++ b/test/cache_invalidation_test.rb
@@ -46,14 +46,14 @@ class CacheInvalidationTest < IdentityCache::TestCase
   def test_after_a_reload_the_cache_perform_as_expected
     Item.cache_has_many :associated_records, :embed => :ids
 
-    assert_equal [@baz, @bar], @record.associated_records
     assert_equal [@baz, @bar], @record.fetch_associated_records
+    assert_equal [@baz, @bar], @record.associated_records
 
     @baz.destroy
     @record.reload
 
-    assert_equal [@bar], @record.associated_records
     assert_equal [@bar], @record.fetch_associated_records
+    assert_equal [@bar], @record.associated_records
   end
 
   def test_cache_invalidation_expire_properly_if_child_is_embed_in_multiple_parents
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 #### 0.3.2
 
+- Use loaded association first when fetching a cache_has_many id embedded association
 - Deprecate setting the inverse active record association on cache hits. Set IdentityCache.never_set_inverse_association to true to avoid this. (#279)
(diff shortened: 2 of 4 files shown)


test/cache_invalidation_test.rb
```diff
@@ -10,6 +10,7 @@ def setup
     @record.save
     @record.reload
     @baz, @bar = @record.associated_records[0], @record.associated_records[1]
+    @record.reload
```
csfrancis: Why did you need to add this `reload`?

csfrancis:
Just the one test question, but code LGTM.

test/cache_invalidation_test.rb
dylanahsmith: The above line loads the association, so `fetch_associated_records` on test_reload_invalidate_cached_objects wasn't actually loading the association from the cache and failed on the assertion for the value in the instance variable that holds the cached association.

test/cache_invalidation_test.rb
```diff
@@ -46,14 +46,14 @@ def test_reload_invalidate_cached_objects
   def test_after_a_reload_the_cache_perform_as_expected
     Item.cache_has_many :associated_records, :embed => :ids
 
-    assert_equal [@baz, @bar], @record.associated_records
     assert_equal [@baz, @bar], @record.fetch_associated_records
+    assert_equal [@baz, @bar], @record.associated_records
```
daniellaniyo: Does performing this assert `assert_equal [@baz, @bar], @record.associated_records` change anything in this case? Or was it to be consistent with the ordering of the asserts that are performed after the reload?

dylanahsmith: Calling `@record.associated_records` first causes the active record association to be loaded, after which `@record.fetch_associated_records` will just return the active record association, which doesn't really test the cache reloading.

Actually, now that I think about it, I'm not sure why the `assert_equal [@baz, @bar], @record.associated_records` lines are needed at all.  We don't need to test active record here.

daniellaniyo: The key piece here would be `@record.reload` and then checking the result returned by `@record.fetch_associated_records` right?

dylanahsmith: The key is that this test should catch regressions in [lib/identity_cache/cache_invalidation.rb](https://github.com/Shopify/identity_cache/blob/v0.3.1/lib/identity_cache/cache_invalidation.rb), but calling `@record.associated_records` before `@record.fetch_associated_records` would miss regressions there because then `@record.fetch_associated_records` would just delegate to `@record.associated_records`

Right, I just don't want to have `@record.fetch_associated_records` delegate to

daniellaniyo:
LGTM

dylanahsmith:
> Also, if the active record association was modified, then the stale cached association would get returned.

It looks like this is also the case for embedded associations, where we use the cached association if it has been loaded, even if the active record association has been loaded.  That should also be fixed, just not as a part of this PR.

We do the right thing on cache_belongs_to associations, where we use the active record association if it is been loaded.

PR #281, opened 2016-07-23
Add a note about fetch_multi

sgringwe:
I'm not sure if this was purposely left out, but this was something nice to find support for and thought it shoudl be added to readme.


camilo:
hmm I probably missed this like 3 years ago. 👍

PR #282, opened 2016-07-28
Mark IDC fetched records as read_only

daniellaniyo:
Adds new configurable behaviour where returned IDC fetched records are read only.

@dylanahsmith @fbogsany for review/thoughts. Cc @camilo

## Problem
There was a PR (#85) that attempted to mark IDC fetched records as read only. However, it was never merged since it required Shopify to be updated in order to work with new behaviour. This PR addresses this issue and allows to have Identity Cache do the right thing by default on the next release with breaking changes. 

Plus, records returned from IDC when `identityCache.should_use_cache?` returns `false - i.e. existing open transaction - are now not marked as read-only. This allows us to keep sharing code to use for reading and writing so that operations stay consistent.

## Solution
Resolves issue #274 by adding a configurable behaviour where IDC returns readonly records without introducing breaking changes. `IdentityCache.fetch_read_only_records` should be set to get the correct behaviour. In this case, returned fetched records will be read-only except for cases where `IdentityCache.should_use_cache?` is `false`.

A first test was done on a shopify branch to confirm that tests pass with these changes [[CI build]](https://buildkite.com/shopify/shopify-branches/builds/135780). `IdentityCache.fetch_read_only_records` will be progressively used to correct instances where we depend on the deprecated behaviour.


lib/identity_cache/belongs_to_caching.rb
```diff
@@ -51,5 +52,15 @@ def #{options[:prepopulate_method_name]}(record)
         CODE
       end
     end
+
+    def load_and_readonlyify(association_name)
```
dylanahsmith: This is also used in the QueryApi, which shouldn't depend on the BelongsToCache module.  The QueryApi module would be a more suitable place for this method.

lib/identity_cache/query_api.rb
```diff
@@ -229,6 +230,7 @@ def require_if_necessary #:nodoc:
       def resolve_cache_miss(id)
         record = self.includes(cache_fetch_includes).reorder(nil).where(primary_key => id).first
         preload_id_embedded_associations([record]) if record
+        record.try(:readonly!) if IdentityCache.fetch_read_only_records
```
dylanahsmith: I dislike `try` because it is poorly documented and has changed behaviour since early versions of rails.  It effectively does `record.respond_to?(:readonly!) ? record : nil` which subtly stops working if the `readonly!` method is removed or renamed.  Usually `try!` is more appropriate, which effectively does `record.nil? ? nil : nil`

In this case I would actually wrap this line and the above line in a `if record` condition.

I think we want to also avoid marking this record as readonly in a transaction

```ruby
    if record
        preload_id_embedded_associations([record])
        record.readonly! if IdentityCache.fetch_read_only_records && IdentityCache.should_use_cache?
    end
```

lib/identity_cache/query_api.rb
```diff
@@ -305,6 +307,7 @@ def find_batch(ids)
         @id_column ||= columns.detect {|c| c.name == primary_key}
         ids = ids.map{ |id| connection.type_cast(id, @id_column) }
         records = where(primary_key => ids).includes(cache_fetch_includes).to_a
+        records.map(&:readonly!) if IdentityCache.fetch_read_only_records && IdentityCache.should_use_cache?
```
dylanahsmith: The result if the map isn't used, so I would use each instead: `records.each(&:readonly!)`

lib/identity_cache/query_api.rb
```diff
@@ -476,5 +484,15 @@ def was_new_record? # :nodoc:
       pk = self.class.primary_key
       !destroyed? && transaction_changed_attributes.has_key?(pk) && transaction_changed_attributes[pk].nil?
     end
+
+    def load_and_readonlyify(association_name)
+      record_or_records = send(association_name)
+      if self.class.reflect_on_association(association_name).collection?
+        record_or_records.map(&:readonly!)
+      else
+        record_or_records.try(:readonly!)
```
dylanahsmith: This is marking records as read-only in the active record association, which could end up having a surprising and annoying side-effect when the association is ever accessed without using IDC.  Accessing the active record association directly shouldn't ever be returning a record from the cache, so it would be safe to modify and save it, but this would break code that does that.

E.g. this doesn't seem like a desirable behaviour

```ruby
item = Item.find(item_id)
uncached_record = item.associated_record
assert_equal false, uncached_record.readonly?
item.fetch_associated_record
assert_equal true, uncached_record.readonly?
```

daniellaniyo: @dylanahsmith 
> so it would be safe to modify and save it

Are you referring to the record from the cache here or the active record association?

dylanahsmith: This method isn't ever for records loaded from the cache.  I think the only purpose is to try to make code that uses `fetch_#{association}` fail deterministically if the records could have been loaded from the cache.  However, in some cases it is marking records as read-only that could have been deterministically loaded from the database, my example tries to show.

lib/identity_cache/query_api.rb
```diff
@@ -183,6 +184,7 @@ def get_embedded_association(record, association, options) #:nodoc:
 
       def coder_from_record(record) #:nodoc:
         unless record.nil?
+          record.readonly! if IdentityCache.fetch_read_only_records && IdentityCache.should_use_cache?
```
dylanahsmith: I don't think we should have to check `IdentityCache.should_use_cache?` here, since I think this method is only used to serialize a record to store it in the cache.

test/denormalized_has_many_test.rb
```diff
@@ -141,6 +141,55 @@ def test_deprecated_set_inverse_association_on_cache_hit
...
+    end
+  end
+
+  def test_db_returned_records_should_never_be_readonly
+    IdentityCache.with_fetch_read_only_records do
+      record_from_db = Item.find(@record.id)
+
+      uncached_records = record_from_db.associated_records
+      uncached_records.each do |record|
+        assert !record.readonly?, "Record #{record} was readonly"
```
dylanahsmith: This isn't really testing anything identity cache, since `Item.find` or `record_from_db.associated_records` are active record methods.  Perhaps you want Item.fetch instead to test a cache miss

daniellaniyo: The intention here was to test the before and after results of `record_from_db.fetch_associated_records` since with the previous `load_and_readonlyify`, `uncached_records.readonly` became true. It asserts that what you mentionned in https://github.com/Shopify/identity_cache/pull/282/files/5647e89b96ff6edcb52a6e3c5a2fd759fb875322#r72793608 is not happening anymore.

Specific tests for cache_hit and miss behavior when read-only is enabled are here: https://github.com/Shopify/identity_cache/blob/df2f676fbccdacb0adbce77709b944e807ed7d36/test/denormalized_has_many_test.rb#L144-L163

test/normalized_belongs_to_test.rb
```diff
@@ -60,4 +60,28 @@ def test_fetching_polymorphic_belongs_to_association
...
+      Item.fetch(@parent_record.id) # warm cache
+      assert @record.fetch_item.readonly?
+      assert_equal false, @record.item.readonly?
+    end
+  end
+
+  def test_returned_record_should_be_readonly_on_cache_miss
+    IdentityCache.with_fetch_read_only_records do
+      assert @record.fetch_item.readonly?
+      assert_equal false, @record.item.readonly?
```
fbogsany: All the `assert false, foo` or `assert foo, false` should be `refute foo`.

test/fetch_test.rb
```diff
@@ -212,4 +212,34 @@ def test_fetch_on_derived_model_raises
...
+    end
+  end
+
+  def test_returned_records_are_not_readonly_with_open_transactions
+    IdentityCache.with_fetch_read_only_records do
+      Item.expects(:fetch).with(1).returns(@record)
+
+      @record.transaction do
+        assert_equal IdentityCache.should_use_cache?, false
+        assert_equal Item.fetch(1).readonly?, false
```
fbogsany: Isn't this just testing the mock?

test/fetch_multi_test.rb
```diff
@@ -223,6 +223,49 @@ def test_fetch_multi_on_derived_model_raises
...
+    IdentityCache.with_fetch_read_only_records do
+      cache_response = {}
+      cache_response[@bob_blob_key] = cache_response_for(@bob)
+      cache_response[@joe_blob_key] = nil
+      cache_response[@fred_blob_key] = cache_response_for(@fred)
+      fetch_multi = fetch_multi_stub(cache_response)
+      assert_equal [@bob, @joe, @fred], Item.fetch_multi(@bob.id, @joe.id, @fred.id)
+      assert fetch_multi.has_been_called_with?(@bob_blob_key, @joe_blob_key, @fred_blob_key)
+
+      Item.fetch_multi(@bob.id, @joe.id, @fred.id).each do |record|
```
fbogsany: While I understand that `fetch_multi_stub` means we're not actually caching anything, calling `Item.fetch_multi` twice in the testcase is confusing. Naively, I'd expect the first call to cache the 3 records (a mix of hits and misses) and the second to return the cached records (all hits). Can you instead do something like:
```ruby
response = Item.fetch_multi(@bob.id, @joe.id, @fred.id)
assert_equal [@bob, @joe, @fred], response
...
response.each do |record|
...
```

test/fetch_multi_test.rb
```diff
@@ -223,6 +223,49 @@ def test_fetch_multi_on_derived_model_raises
...
+    IdentityCache.with_fetch_read_only_records do
+      cache_response = {}
+      cache_response[@bob_blob_key] = nil
+      cache_response[@joe_blob_key] = nil
+      cache_response[@fred_blob_key] = cache_response_for(@fred)
+      fetch_multi = fetch_multi_stub(cache_response)
+      assert_equal [@bob, @joe, @fred], Item.fetch_multi(@bob.id, @joe.id, @fred.id)
+      assert fetch_multi.has_been_called_with?(@bob_blob_key, @joe_blob_key, @fred_blob_key)
+
+      Item.fetch_multi(@bob.id, @joe.id, @fred.id).each do |record|
```
fbogsany: Same comment as above.

test/fetch_test.rb
daniellaniyo: 🙆 totally right! I'll spy call_through

test/fetch_multi_test.rb
```diff
@@ -223,6 +223,49 @@ def test_fetch_multi_on_derived_model_raises
...
+        assert record.readonly?, "Record #{record} wasn't readonly"
+      end
+    end
+  end
+
+  def test_fetch_multi_with_open_transactions_returns_non_readonly_records
+    IdentityCache.with_fetch_read_only_records do
+      Item.connection.expects(:open_transactions).at_least_once.returns(1)
```
fbogsany: I'm not sure whether it's just historical, but why do we use this expectation here instead of an actual `Item.transaction` block? We're using the transaction block in similar tests, e.g. in `test/normalized_belongs_to_test.rb`.

test/fetch_multi_test.rb
```diff
@@ -223,6 +223,49 @@ def test_fetch_multi_on_derived_model_raises
...
+      assert_equal IdentityCache.should_use_cache?, false
+      IdentityCache.cache.expects(:fetch_multi).never
+      Item.fetch_multi(@bob.id, @joe.id, @fred.id).each do |record|
+        assert !record.readonly?, "Record #{record} was readonly"
```
fbogsany: `refute record.readonly?, ...`

test/denormalized_has_one_test.rb
```diff
@@ -120,4 +120,45 @@ def test_cache_has_one_on_derived_model_raises
       StiRecordTypeA.cache_has_one :polymorphic_record, :inverse_name => :owner, :embed => true
     end
   end
+
+  def test_returned_record_should_be_readonly_on_cache_hit
+    IdentityCache.with_fetch_read_only_records do
+      Item.fetch_by_title('foo')
```
fbogsany: Why not `Item.fetch(1)` here? That'd make it clearer that the subsequent `fetch` is a cache hit.

lib/identity_cache/query_api.rb
```diff
@@ -183,6 +184,7 @@ def get_embedded_association(record, association, options) #:nodoc:
 
       def coder_from_record(record) #:nodoc:
         unless record.nil?
+          record.readonly! if IdentityCache.fetch_read_only_records
```
fbogsany: Why are we doing this here? Because it's a useful bottleneck interface? (I.e. only all relevant paths go through here.) It seems wrong to modify the `record` argument in a method called `coder_from_record`.

test/denormalized_has_many_test.rb
```diff
@@ -141,6 +141,55 @@ def test_deprecated_set_inverse_association_on_cache_hit
...
   end
 
+  def test_returned_records_should_be_readonly_on_cache_hit
+    IdentityCache.with_fetch_read_only_records do
+      Item.fetch(@record.id) # warm cache
+
+      record_from_cache_hit = Item.fetch(@record.id)
+      record_from_cache_hit.fetch_associated_records.each do |record|
+        assert record.readonly?, "Record #{record} wasn't readonly"
+      end
```
fbogsany: While the clarity of the error message on test failure is nice, I would probably use the more concise:
```ruby
assert record_from_cache_hit.fetch_associated_records.all? &:readonly?
```

lib/identity_cache/query_api.rb
daniellaniyo: @fbogsany: In one of the previous commit, we were calling `load_and_readonlyify` to mark records as read-only at a higher level such as [`fetch_recursively_cached_association`](https://github.com/Shopify/identity_cache/pull/282/commits/5647e89b96ff6edcb52a6e3c5a2fd759fb875322#diff-17e7d279c636296c9b82ebbef85db611R429). However, the issue was that when calling a `fetch_associated_records` on an item deterministically pulled from the db,  `item.associated_record` was being marked as read-only. The idea here - not ideal though - was to mark them as read-only at a lower level, while preserving the integrity of the records pulled deterministically from the db.

The other alternative that @dylanahsmith suggested was to duplicate records and mark those as read-only.

test/fetch_multi_test.rb
daniellaniyo: From looking at the other tests and the historical timeline, `fetch_multi_test` is one of the oldest tests and was written with the `open_transactions` stub which stuck around. I can update it and use `Item.transaction` everywhere in `fetch_multi_test.rb`.

test/normalized_belongs_to_test.rb
dylanahsmith: I usually prefer `assert_equal false, ` over `refute` because it is more specific, however, I don't care as much here since `readonly?` is a rails method.

daniellaniyo:
@fbogsany  Thanks for your review! I refactored tests based on our feedback. Please see the latest commit.

daniellaniyo:
Thank you both for your feedback! For [the last commit](https://github.com/Shopify/identity_cache/pull/282/commits/420989ab2a8009adce693ec3c1b4f75e7962a415),  I mainly moved the record readonly-ifying operation from `coder-from-record` to `fetch_recursively_cached_association` that set the `cached#{association}` records. Records loaded from the db are first duplicated and then marked as read-only to preserve records that could have been deterministically loaded from the db.

@dylanahsmith @fbogsany could you please review?

test/denormalized_has_many_test.rb
```diff
@@ -141,6 +141,40 @@ def test_deprecated_set_inverse_association_on_cache_hit
...
+      record_from_cache_miss = Item.fetch(@record.id)
+      assert record_from_cache_miss.fetch_associated_records.all?(&:readonly?)
+    end
+  end
+
+  def test_db_returned_records_should_never_be_readonly
+    IdentityCache.with_fetch_read_only_records do
+      record_from_db = Item.find(@record.id)
+      uncached_records = record_from_db.associated_records
+      refute uncached_records.all?(&:readonly?)
```
dylanahsmith: I would use `assert uncached_records.none?(&:readonly?)` so this doesn't pass if some but not all records are readonly.

lib/identity_cache/query_api.rb
```diff
@@ -476,5 +486,16 @@ def was_new_record? # :nodoc:
...
+
+    def load_and_readonlyify(association_name)
+      record_or_records = send(association_name)
+
+      readonly_record_or_records = if self.class.reflect_on_association(association_name).collection?
+        record_or_records.map { |p| p.dup.tap(&:readonly!) }
+      else
+        record_or_records.dup.tap(&:readonly!) if record_or_records
+      end
+      readonly_record_or_records
```
dylanahsmith: You don't need this variable.  You can just end the function on the if statement to return its result

test/denormalized_has_many_test.rb
```diff
@@ -141,6 +141,40 @@ def test_deprecated_set_inverse_association_on_cache_hit
...
+      record_from_db.fetch_associated_records
+      refute uncached_records.all?(&:readonly?)
```
dylanahsmith: This isn't testing the result of the above line.  Maybe just write it as `assert record_from_db.fetch_associated_records.none?(&:readonly?)` to actually test `fetch_associated_records`.

After that, this test will fail because `fetch_recursively_cached_association` calls `load_and_readonlyify` on the records.

test/denormalized_has_one_test.rb
dylanahsmith: Why would that make it clearer?  I think it is clearer the way it is, since it is doing the same thing twice, and using a variable to indicate that the following fetch is a cache hit.  Also, `fetch_by_#{attribute}` will use an extra layer of caching to lookup the record's id before doing `Item.fetch(1)`.

dylanahsmith: Using `Item.fetch(1)` on this line and the following line would make sense though, since the focus on this test isn't on `fetch_by_title`, so we may as well skip that layer of caching

test/denormalized_has_many_test.rb
daniellaniyo: The idea of the test was to see if the `fetch_associated_records` affects the original records returned from the db. Given this context, shouldn't the checks be `assert record_from_db.fetch_associated_records.all?(&:readonly?)` and then `assert record_from_db.associated_records.none?(&:readonly?)`?

dylanahsmith: Oh, I see.  I had a different idea of what would be marked read-only.  Since `record_from_db.fetch_associated_records` wouldn't ever fetch records from the cache here, I figured the records don't need to be marked read-only.  But since you aren't marking the active record association's records as read-only, then this should be :ok:

dylanahsmith:
:+1:

daniellaniyo:
Squashing and posting performance tests results.

daniellaniyo:
TL; DR: No significant difference were observed; reported performance/loss are within 10% with the highest delta (highest loss) being 0.1119% for `FetchEmbedDeletedConflictRunner`

IDC CPU Benchmark tests results:

**On master:**
```
$ dev benchmark-cpu
                                            user     system      total        real
FindRunner:                             0.880000   0.060000   0.940000 (  2.883850)
FetchEmbedMissRunner:                   2.290000   0.070000   2.360000 (  4.430225)
FetchEmbedHitRunner:                    2.320000   0.080000   2.400000 (  4.471694)
FetchEmbedDeletedRunner:                2.340000   0.070000   2.410000 (  4.539874)
FetchEmbedConflictRunner:               2.570000   0.070000   2.640000 (  4.740887)
FetchEmbedDeletedConflictRunner:        2.600000   0.080000   2.680000 (  4.772136)
FetchNormalizedMissRunner:              3.180000   0.100000   3.280000 (  5.317594)
FetchNormalizedHitRunner:               3.270000   0.100000   3.370000 (  5.433163)
FetchNormalizedDeletedRunner:           3.240000   0.100000   3.340000 (  5.411182)
FetchNormalizedConflictRunner:          2.700000   0.080000   2.780000 (  4.866313)
FetchNormalizedDeletedConflictRunner:   2.480000   0.090000   2.570000 (  4.640315)
```

**On branch:**
```
$ dev benchmark-cpu
                                            user     system      total        real
FindRunner:                             0.870000   0.060000   0.930000 (  2.900611)
FetchEmbedMissRunner:                   2.300000   0.080000   2.380000 (  4.448018)
FetchEmbedHitRunner:                    2.340000   0.070000   2.410000 (  4.479704)
FetchEmbedDeletedRunner:                2.400000   0.080000   2.480000 (  4.560004)
FetchEmbedConflictRunner:               2.580000   0.070000   2.650000 (  4.754499)
FetchEmbedDeletedConflictRunner:        2.630000   0.080000   2.710000 (  4.835958)
FetchNormalizedMissRunner:              3.180000   0.100000   3.280000 (  5.309665)
FetchNormalizedHitRunner:               3.200000   0.100000   3.300000 (  5.325748)
FetchNormalizedDeletedRunner:           2.960000   0.110000   3.070000 (  5.102926)
FetchNormalizedConflictRunner:          2.430000   0.080000   2.510000 (  4.596064)
FetchNormalizedDeletedConflictRunner:   2.440000   0.080000   2.520000 (  4.611049)
```

**Performance gain/loss** 
(_computed using the above total values in (master_value-branch_value)-master_value_ )
```
FindRunner:                            -0.01063829787
FetchEmbedMissRunner:                  0.008474576271
FetchEmbedHitRunner:                   0.004166666667
FetchEmbedDeletedRunner:               0.02904564315
FetchEmbedConflictRunner:              0.003787878788
FetchEmbedDeletedConflictRunner:       0.01119402985
FetchNormalizedMissRunner:             0
FetchNormalizedHitRunner:              -0.02077151335
FetchNormalizedDeletedRunner:          -0.08083832335
FetchNormalizedConflictRunner:         -0.09712230216
FetchNormalizedDeletedConflictRunner:  -0.01945525292
```

Profile tests follow.

test/fetch_test.rb
```diff
@@ -212,4 +212,34 @@ def test_fetch_on_derived_model_raises
...
+      assert fetch.has_been_called_with?(@blob_key)
+      assert Item.fetch(1).readonly?
+    end
+  end
+
+  def test_returned_records_are_not_readonly_with_open_transactions
+    IdentityCache.with_fetch_read_only_records do
+
+      @record.transaction do
+        fetch = Spy.on(IdentityCache.cache, :fetch).and_call_through
```
fbogsany: I might need more ☕ , but I can't see where this `fetch` is used. AFAICT this line isn't necessary.

test/fetch_multi_test.rb
```diff
@@ -223,6 +223,48 @@ def test_fetch_multi_on_derived_model_raises
...
+      assert response.all?(&:readonly?)
+    end
+  end
+
+  def test_fetch_multi_with_open_transactions_returns_non_readonly_records
+    IdentityCache.with_fetch_read_only_records do
+      Item.transaction do
+        assert_equal IdentityCache.should_use_cache?, false
+        IdentityCache.cache.expects(:fetch_multi).never
+        refute Item.fetch_multi(@bob.id, @joe.id, @fred.id).all?(&:readonly?)
```
fbogsany: I think you want `assert ...none?(&:readonly?)` here.

test/denormalized_has_many_test.rb
```diff
@@ -141,6 +141,40 @@ def test_deprecated_set_inverse_association_on_cache_hit
...
+      assert record_from_db.fetch_associated_records.all?(&:readonly?)
+      assert record_from_db.associated_records.none?(&:readonly?)
+    end
+  end
+
+  def test_returned_records_with_open_transactions_should_not_be_readonly
+    IdentityCache.with_fetch_read_only_records do
+      Item.transaction do
+        assert_equal IdentityCache.should_use_cache?, false
+        refute Item.fetch(@record.id).fetch_associated_records.all?(&:readonly?)
```
fbogsany: I think this should be `assert ...none?(&:readonly?)`.

daniellaniyo:
Profile for `FetchEmbedDeletedonflictRunner`:

**On master**
```
FetchEmbedDeletedConflictRunner:
==================================
  Mode: cpu(1000)
  Samples: 3348 (0.00% miss rate)
  GC: 195 (5.82%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
      2761  (82.5%)        2761  (82.5%)     block in Mysql2::Client#query
        54   (1.6%)          54   (1.6%)     Memcached::Error#set_backtrace
      2787  (83.2%)          22   (0.7%)     ActiveRecord::ConnectionAdapters::Mysql2Adapter#exec_query
       106   (3.2%)          15   (0.4%)     Memcached#check_return_code
        14   (0.4%)          14   (0.4%)     ActiveSupport::PerThreadRegistry#instance
        13   (0.4%)          13   (0.4%)     ActiveSupport::Cache::Entry#initialize
         9   (0.3%)           9   (0.3%)     ActiveRecord::Relation#reset
         9   (0.3%)           9   (0.3%)     block in Memcached#server_structs
         8   (0.2%)           8   (0.2%)     Time#compare_with_coercion
         8   (0.2%)           8   (0.2%)     block (4 levels) in Class#class_attribute
        11   (0.3%)           8   (0.2%)     #<Module:0x007fbcbaa98f20>.encode
         7   (0.2%)           7   (0.2%)     ActiveRecord::QueryMethods#order_values
         7   (0.2%)           7   (0.2%)     block in ActiveRecord::QueryMethods#validate_order_args
        15   (0.4%)           6   (0.2%)     ActiveRecord::Relation#initialize_copy
        14   (0.4%)           6   (0.2%)     block in ActiveRecord::Relation::Merger#merge
         5   (0.1%)           5   (0.1%)     ActiveRecord::QueryMethods#where_values
        14   (0.4%)           5   (0.1%)     ActiveRecord::QueryMethods#preprocess_order_args
         5   (0.1%)           5   (0.1%)     ActiveRecord::Relation#values
        16   (0.5%)           5   (0.1%)     ActiveModel::AttributeMethods::ClassMethods#attribute_alias?
         4   (0.1%)           4   (0.1%)     ActiveRecord::Inheritance::ClassMethods#base_class
```


**On `support_fetch_readonly_records` branch**
```
FetchEmbedDeletedConflictRunner:
==================================
  Mode: cpu(1000)
  Samples: 3526 (0.00% miss rate)
  GC: 177 (5.02%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
      2688  (76.2%)        2688  (76.2%)     block in Mysql2::Client#query
        91   (2.6%)          91   (2.6%)     Memcached::Error#set_backtrace
      2726  (77.3%)          26   (0.7%)     ActiveRecord::ConnectionAdapters::Mysql2Adapter#exec_query
        24   (0.7%)          24   (0.7%)     ActiveSupport::PerThreadRegistry#instance
       158   (4.5%)          20   (0.6%)     Memcached#check_return_code
        23   (0.7%)          18   (0.5%)     #<Module:0x007fee19270c40>.encode
        42   (1.2%)          17   (0.5%)     ActiveModel::AttributeMethods::ClassMethods#attribute_alias?
        17   (0.5%)          17   (0.5%)     ThreadSafe::NonConcurrentCacheBackend#[]
        36   (1.0%)          16   (0.5%)     ActiveRecord::QueryMethods#preprocess_order_args
        16   (0.5%)          16   (0.5%)     block (4 levels) in Class#class_attribute
        14   (0.4%)          14   (0.4%)     Time#compare_with_coercion
        14   (0.4%)          13   (0.4%)     ActiveSupport::Cache::Entry#initialize
        19   (0.5%)          12   (0.3%)     ActiveRecord::LazyAttributeHash#[]
        16   (0.5%)          10   (0.3%)     ActiveRecord::Relation#initialize_copy
        10   (0.3%)          10   (0.3%)     ActiveRecord::Core::ClassMethods#arel_table
       138   (3.9%)           9   (0.3%)     Memcached#reraise
         9   (0.3%)           9   (0.3%)     block in ActiveRecord::QueryMethods#preprocess_order_args
         9   (0.3%)           9   (0.3%)     ActiveRecord::Inheritance::ClassMethods#base_class
        14   (0.4%)           8   (0.2%)     IdentityCache::QueryAPI::ClassMethods#all_cached_associations
        28   (0.8%)           8   (0.2%)     block in ActiveRecord::Relation::Merger#merge
```

More time spent for Memcached::Error#set_backtrace; however this seems unrelated to the suggested changes.  Other increase in cpu % (+0.3%) are seen with`ActiveSupport::PerThreadRegistry#instance`,   `Memcached#check_return_code` that are external to IDC.   We also do notice a 0.2% for `IdentityCache::QueryAPI::ClassMethods#all_cached_associations` that wasn't present in the profile results from master.

test/fetch_test.rb
daniellaniyo: There was a `refute fetch.has_been_called_with?(@blob_key)` check that I unexpectedly forgot. I'm the one who needs ☕

fbogsany:
LGTM 👍

dylanahsmith:
:shipit:

PR #283, opened 2016-07-28
Remove support for rails 4.0 and 4.1 since they aren't supported upsteam

dylanahsmith:
@rafaelfranca & @fbogsany for review

It looks like only rails 4.2 and 5.0 are supported (http://guides.rubyonrails.org/maintenance_policy.html) so we should drop support for rails versions that aren't even supported upstream for severe security issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,6 @@ rvm:
   - 2.2.3
 
 gemfile:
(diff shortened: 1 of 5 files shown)


rafaelfranca:
:shipit:

PR #285, opened 2016-07-29
Fix tests for rails 5.0 with mysql and add it to CI

dylanahsmith:
@sgrif & @rafaelfranca for review

CI wasn't running tests for Rails 5.0, so there are some failures.  Most of the failures just required some tests to be updated.

There is still a test failure for rails 5.0 with postgresql which I haven't had the time to dig into yet, so decided to exclude it for now so we can get CI coverage for rails 5.0 mysql.  The failure was for the serialization format, which should be independent of rails now, but apparently isn't.  We are using `attributes_before_type_cast` and the same version of the pg gem with rails 5.0, but the type of the "id" values changed from a string to an integer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,14 @@ rvm:
 
 gemfile:
   - Gemfile.rails42
+  - Gemfile.rails50
+
+matrix:
+  exclude:
+    - rvm: 2.1
+      gemfile: Gemfile.rails50
+    - gemfile: Gemfile.rails50
+      env: DB=postgresql
 
 env:
   - DB=mysql2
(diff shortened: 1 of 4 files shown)


rafaelfranca:
LGTM.

The postgresql related issue seems correct. If the "id" that you are referring the value after loaded or the value stored in the cache?

dylanahsmith:
> If the "id" that you are referring the value after loaded or the value stored in the cache?

I'm referring to the "id" value that we store in the cache, where we try to store the attributes before type casting to make the cache format rails independent.  If changing from rails 4.2 to 5.0 is changing the serialization format, then it doesn't seem rails independent.

rafaelfranca:
I'd expect that to change when reading from the cache, but the value in the cache should not have changed. @sgrif do you remember anything about this?

dylanahsmith:
The rails 5 postgresql issue can be addressed in a follow-up, so I'm merging this.

PR #287, opened 2016-08-09
New version 0.4.0 - Fetch association returns an array

daniellaniyo:
@fbogsany & @dylanahsmith for review. Cc @camilo 

Bumping the gem version to 0.4.0. Fetch association no longer returns a relation; but rather an array. This will prevent chaining in the future. 

This PR follows https://github.com/Shopify/identity_cache/pull/276 where the `fetch_returns_relation` behavior was made configurable. Also note that this should not cause any break on the shopify side since core is already running with `IdentityCache.fetch_return_relation=false`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # IdentityCache changelog
 
+#### 0.4.0
+
+- Return an array from fetched association to prevent chaining. Up to now, a relation was returned by default. (#287)
+
 #### 0.3.2
 
 - Deprecate returning non read-only records when cache is used. Set IdentityCache.fetch_readonly_records to true to avoid this. (#282)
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # IdentityCache changelog
 
+#### 0.4.0
+
+- Return an array from fetched association to prevent chaining. Up to now, a relation was returned by default. (#287)
+
 #### 0.3.2
```
dylanahsmith: Could we first make a 0.3.2 release?  Currently the latest on [rubygems](https://rubygems.org/gems/identity_cache) is 0.3.1

dylanahsmith:
:+1:

CHANGELOG.md
daniellaniyo: Already pushed the v.0.3.2 tag. Shipit fails to pick up the tag. We might need to use the traditonal `rake release` while the issue is being debugged.

daniellaniyo: Done! 0.3.2 released

fbogsany:
👍

PR #288, opened 2016-08-11
Consistently store the array for cached has many associations

dylanahsmith:
@daniellaniyo, @rafaelfranca for review
cc @csfrancis, @camilo

## Problem

This is a second try at solving the inconsistency that #277 attempted to address.  However, that PR was reverted because it was instantiating the collection proxy instances when associations were loaded from the cache rather than lazily when they were accessed, which was causing a significant performance impact.

## Solution

Thanks to @daniellaniyo's work with making `fetch_#{association_name}` cache has many methods return an array instead of a collection proxy, we no longer need to store the cached association as a collection proxy in the instance variable.  Now we can just store an array of records loaded from the cache.

Also, we can get a performance improvement by using `association(association_name).load_target` to load the array using the association without instantiating a collection proxy object.

### master

```
FindRunner:                             0.850000   0.040000   0.890000 (  1.715479)
FetchEmbedMissRunner:                   2.340000   0.060000   2.400000 (  3.328042)
FetchEmbedHitRunner:                    2.300000   0.060000   2.360000 (  3.280951)
FetchEmbedDeletedRunner:                2.300000   0.060000   2.360000 (  3.287030)
FetchEmbedConflictRunner:               2.520000   0.060000   2.580000 (  3.531867)
FetchEmbedDeletedConflictRunner:        2.520000   0.060000   2.580000 (  3.514359)
FetchNormalizedMissRunner:              3.050000   0.080000   3.130000 (  4.109704)
FetchNormalizedHitRunner:               3.060000   0.080000   3.140000 (  4.106677)
FetchNormalizedDeletedRunner:           2.960000   0.070000   3.030000 (  3.972951)
FetchNormalizedConflictRunner:          2.440000   0.060000   2.500000 (  3.408446)
FetchNormalizedDeletedConflictRunner:   2.430000   0.050000   2.480000 (  3.408808)
```

### this branch

```
FindRunner:                             0.800000   0.040000   0.840000 (  1.639428)
FetchEmbedMissRunner:                   1.540000   0.050000   1.590000 (  2.462227)
FetchEmbedHitRunner:                    1.520000   0.050000   1.570000 (  2.447718)
FetchEmbedDeletedRunner:                1.540000   0.060000   1.600000 (  2.457482)
FetchEmbedConflictRunner:               1.710000   0.050000   1.760000 (  2.644591)
FetchEmbedDeletedConflictRunner:        1.710000   0.060000   1.770000 (  2.628639)
FetchNormalizedMissRunner:              2.790000   0.070000   2.860000 (  3.775260)
FetchNormalizedHitRunner:               2.800000   0.080000   2.880000 (  3.767972)
FetchNormalizedDeletedRunner:           2.810000   0.070000   2.880000 (  3.782054)
FetchNormalizedConflictRunner:          1.710000   0.050000   1.760000 (  2.621399)
FetchNormalizedDeletedConflictRunner:   1.710000   0.050000   1.760000 (  2.628674)
```

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -152,16 +152,17 @@ module IdentityCache
         value = if IdentityCache.unmap_cached_nil_for(coder_or_array).nil?
           nil
         elsif (reflection = record.class.reflect_on_association(association_name)).collection?
-          association = reflection.association_class.new(record, reflection)
-          association.target = coder_or_array.map {|e| record_from_coder(e) }
+          associated_records = coder_or_array.map {|e| record_from_coder(e) }
 
-          set_inverse_of_cached_has_many(record, reflection, association.target)
+          set_inverse_of_cached_has_many(record, reflection, associated_records)
 
           unless IdentityCache.never_set_inverse_association
+            association = reflection.association_class.new(record, reflection)
+            association.target = associated_records
             association.target.each {|e| association.set_inverse_instance(e) }
           end
 
-          association
+          associated_records
         else
           record_from_coder(coder_or_array)
         end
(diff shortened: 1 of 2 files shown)


rafaelfranca:
LGTM. The code is easier to understand now

lib/identity_cache/query_api.rb
```diff
@@ -483,13 +479,17 @@ def was_new_record? # :nodoc:
...
+    end
 
-      if self.class.reflect_on_association(association_name).collection?
-        record_or_records.map { |p| p.dup.tap(&:readonly!) }
-      else
-        record_or_records.dup.tap(&:readonly!) if record_or_records
+    def readonly_copy(record_or_records)
+      if record_or_records.respond_to?(:to_ary)
+        record_or_records.map { |record| readonly_record_copy(record) }
+      elsif record_or_records
```
camilo: why is this check needed? can the record be nil?

lib/identity_cache/query_api.rb
```diff
@@ -483,13 +479,17 @@ def was_new_record? # :nodoc:
...
+      record.readonly!
+      record
+    end
 
-      if self.class.reflect_on_association(association_name).collection?
-        record_or_records.map { |p| p.dup.tap(&:readonly!) }
-      else
-        record_or_records.dup.tap(&:readonly!) if record_or_records
+    def readonly_copy(record_or_records)
+      if record_or_records.respond_to?(:to_ary)
```
camilo: what quacks like that: `to_ary` ?

camilo:
couple questions, besides that 👍

lib/identity_cache/query_api.rb
dylanahsmith: yes, `record_or_records` is the result of loading an association, where a has_one association could return `nil`

lib/identity_cache/query_api.rb
dylanahsmith: This is to handle when it is an array of records.  I guess it would be clearer if I changed it to `if record_or_records.is_a?(Array)`

daniellaniyo:
LGTM and 🙌  for the `load_target` improvement.

PR #291, opened 2016-08-25
Move should_use_cache? calls to the model-level

Sirupsen:
I'd like to be able to override `#should_use_cache?` on specific models to have domain-specific logic to decide whether the cache should be used. This is the minimal change to the code to support such a use-case through method overriding.

The tests are not great, I'm not super familiar with IDC. I can definitely dig more into this.

Please review @dylanahsmith @fbogsany @daniellaniyo

cc @Shopify/pods 

<hr/>

In my case, I have multiple databases in my application. When a particular database connection is to a delayed read-slave, I would like to disable IDC, both reads and writes. While I could continue to perform reads with IDC, my objective is to move off of IDC for the models for this database. Using delayed read-slaves and IDC together is complicated—I'd like to transition to only using the read-slave iteratively. However, before I can disable IDC completely, this is key for me to be able to override as the model may sometimes be used from a readonly context (don't use IDC) and sometimes from a writer context (sure, use IDC).

```ruby
class Shop < ActiveRecord::Base
  include IdentityCache

  def should_use_cache?
     !read_only_connection? && IdentityCache.should_use_cache? 
  end
end

with_read_only_connection do
  Shop.fetch_by_id(12) # will never hit IDC
end

Shop.fetch_by_id(12) # will hit IDC
```

<details>
<summary>**Why did you not do the same for `#should_fill_cache?`.** </summary>
For my use-case for transitioning models off of IDC, it doesn't make sense to continue filling the cache when I am not going to use it eventually. 

Another reason is that if I just disable filling the cache, I will have a situation where after the 4 hour default expiry all the cache entires will expire and my load will be significantly different. Instead of noticing that 4 hours later, I would much rather it be instant while I'm monitoring the rollout.

Even though I don't need it personally, it could make sense to support it, however, the `CacheFetcher` is the only user of this and it doesn't have access to the model. I don't want to start passing down the model or a boolean on every call unnecessarily.
</details>

<details>
  <summary>**Why support this by overriding a public method and not an official API?**</summary>
Instead of providing an interface that allows overriding the method, we could instead expose an API such as:

```ruby
class Shop < ActiveRecord::Base
  include IdentityCache
  self.should_use_cache = -> { !read_only_connection }
end
```

However, this seems like such an obscure use-case to support publicly that I think structuring the code in a way that makes it possible is more important than providing it as a first-class citizen. Especially given the only use-case that has surfaced is for migration purposes.
</details>

### DIFF
--- a/lib/identity_cache/belongs_to_caching.rb
+++ b/lib/identity_cache/belongs_to_caching.rb
@@ -34,14 +34,15 @@ module IdentityCache
         foreign_key = options[:association_reflection].foreign_key
         self.class_eval(<<-CODE, __FILE__, __LINE__ + 1)
           def #{options[:cached_accessor_name]}
-            if IdentityCache.should_use_cache? && #{foreign_key}.present? && !association(:#{association}).loaded?
+            association_klass = association(:#{association}).klass
+            if association_klass.should_use_cache? && #{foreign_key}.present? && !association(:#{association}).loaded?
               if instance_variable_defined?(:@#{options[:records_variable_name]})
                 @#{options[:records_variable_name]}
               else
-                @#{options[:records_variable_name]} = association(:#{association}).klass.fetch_by_id(#{foreign_key})
+                @#{options[:records_variable_name]} = association_klass.fetch_by_id(#{foreign_key})
               end
             else
-              if IdentityCache.fetch_read_only_records && IdentityCache.should_use_cache?
+              if IdentityCache.fetch_read_only_records && association_klass.should_use_cache?
                 load_and_readonlyify(:#{association})
               else
                 #{association}
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -233,7 +233,8 @@ module IdentityCache
           end
 
           def #{options[:cached_accessor_name]}
-            if IdentityCache.should_use_cache? && !#{association}.loaded?
+            association_klass = association(:#{association}).klass
+            if association_klass.should_use_cache? && !#{association}.loaded?
               @#{options[:records_variable_name]} ||= #{options[:association_reflection].klass}.fetch_multi(#{options[:cached_ids_name]})
             else
               #{association}.to_a
@@ -251,13 +252,22 @@ module IdentityCache
 
       def attribute_dynamic_fetcher(attribute, fields, values, unique_index) #:nodoc:
         raise_if_scoped
-        cache_key = rails_cache_key_for_attribute_and_fields_and_values(attribute, fields, values, unique_index)
-        IdentityCache.fetch(cache_key) do
-          query = reorder(nil).where(Hash[fields.zip(values)])
-          query = query.limit(1) if unique_index
-          results = query.pluck(attribute)
-          unique_index ? results.first : results
+
+        if should_use_cache?
+          cache_key = rails_cache_key_for_attribute_and_fields_and_values(attribute, fields, values, unique_index)
+          IdentityCache.fetch(cache_key) do
+            dynamic_attribute_cache_miss(attribute, fields, values, unique_index)
+          end
+        else
+          dynamic_attribute_cache_miss(attribute, fields, values, unique_index)
         end
+      end
+
+      def dynamic_attribute_cache_miss(attribute, fields, values, unique_index)
+        query = reorder(nil).where(Hash[fields.zip(values)])
+        query = query.limit(1) if unique_index
+        results = query.pluck(attribute)
+        unique_index ? results.first : results
       end
 
       def add_parent_expiry_hook(options)
@@ -265,6 +275,7 @@ module IdentityCache
 
         child_class.send(:include, ArTransactionChanges) unless child_class.include?(ArTransactionChanges)
         child_class.send(:include, ParentModelExpiration) unless child_class.include?(ParentModelExpiration)
+        child_class.send(:include, ShouldUseCache) unless child_class.respond_to?(:should_use_cache?)
 
         child_class.parent_expiration_entries[options[:inverse_name]] << [self, options[:only_on_foreign_key_change]]
 
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -21,7 +21,7 @@ module IdentityCache
         raise_if_scoped
         raise NotImplementedError, "fetching needs the primary index enabled" unless primary_cache_index_enabled
         return unless id
-        record = if IdentityCache.should_use_cache?
+        record = if should_use_cache?
           require_if_necessary do
             object = nil
             coder = IdentityCache.fetch(rails_cache_key(id)){ coder_from_record(object = resolve_cache_miss(id)) }
@@ -53,7 +53,7 @@ module IdentityCache
         raise NotImplementedError, "fetching needs the primary index enabled" unless primary_cache_index_enabled
         options = ids.extract_options!
         ids.flatten!(1)
-        records = if IdentityCache.should_use_cache?
+        records = if should_use_cache?
           require_if_necessary do
             cache_keys = ids.map {|id| rails_cache_key(id) }
             key_to_id_map = Hash[ cache_keys.zip(ids) ]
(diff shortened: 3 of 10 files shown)


dylanahsmith:
Sounds like you are trying to do the same thing as https://github.com/Shopify/identity_cache/pull/256.

Sirupsen:
Oh, I didn't know about that PR, yes I am. Is there an approach you prefer @dylanahsmith?

dylanahsmith:
Took a quick look, and it looks like this PR avoids a point of contention on the other PR (https://github.com/Shopify/identity_cache/pull/256/files#r58312724)

daniellaniyo:
👍

camilo:
As far as I can see this just disables entirely for a model. IMO that is a lot clearer/better than letting the client pick and choose on every call to `fetch`

lib/identity_cache/configuration_dsl.rb
```diff
@@ -233,7 +237,14 @@ def #{options[:cached_ids_name]}
           end
 
           def #{options[:cached_accessor_name]}
-            if IdentityCache.should_use_cache? && !#{association}.loaded?
+            association_klass = association(:#{association}).klass
+            should_use_cache = if association_klass.respond_to?(:should_use_cache?)
```
dylanahsmith: Why is this `respond_to?` check necessary?  This is for cache_has_many with `embed: :ids` where those ids will be loaded using the associated model, thus that model must already `include IdentityCache`

lib/identity_cache/belongs_to_caching.rb
```diff
@@ -34,14 +34,21 @@ def build_normalized_belongs_to_cache(association, options)
         foreign_key = options[:association_reflection].foreign_key
         self.class_eval(<<-CODE, __FILE__, __LINE__ + 1)
           def #{options[:cached_accessor_name]}
-            if IdentityCache.should_use_cache? && #{foreign_key}.present? && !association(:#{association}).loaded?
+            association_klass = association(:#{association}).klass
+            should_use_cache = if association_klass.respond_to?(:should_use_cache?)
```
dylanahsmith: This shouldn't be necessary, since we already need the associated class to `include IdentityCache` to call `fetch_by_id` on it.

lib/identity_cache/query_api.rb
```diff
@@ -79,7 +79,7 @@ def fetch_multi(*ids)
       def prefetch_associations(associations, records)
         records = records.to_a
         return if records.empty?
-        unless IdentityCache.should_use_cache?
+        unless should_use_cache?
```
dylanahsmith: Elsewhere you seem to be doing this check on each cache_has_many association, but here it is done for multiple cache_has_many associations

lib/identity_cache/query_api.rb
```diff
@@ -417,7 +417,14 @@ def prefetch_one_association(association, records)
     private
 
     def fetch_recursively_cached_association(ivar_name, association_name) # :nodoc:
-      assoc = if IdentityCache.should_use_cache?
+      association_klass = association(association_name).klass
+      should_use_cache = if association_klass.respond_to?(:should_use_cache?)
```
dylanahsmith: Should we instead include a module that defines this method [as we do with ParentModelExpiration](https://github.com/Shopify/identity_cache/blob/master/lib/identity_cache/configuration_dsl.rb).  That way we wouldn't need a special case for it here and could just use `associated_klass.should_use_cache?` here.

lib/identity_cache/query_api.rb
```diff
@@ -227,7 +227,7 @@ def resolve_cache_miss(id)
         record = self.includes(cache_fetch_includes).reorder(nil).where(primary_key => id).first
```
dylanahsmith: If any of these embedded associations that are included through `cache_fetch_includes` are read from a read slave, then they will be cached with stale data if `self.should_use_cache?` is true.  This could be a problem if a model that uses one database has a cache_has_many with `embed: true` for an association where the associated model uses another database.

dylanahsmith:
You missed `cache_attribute` which currently relies on `IdentityCache.fetch` to call `IdentityCache.should_use_cache?`, but that won't be done in a model aware manner.

lib/identity_cache/query_api.rb
dylanahsmith: Talked with @Sirupsen about this in person and we plan on banning embedded an association between models that use different databases at the application level.

Sirupsen: 👍

Sirupsen:
@dylanahsmith everything should be good now! I'm going to wait with rebasing so I can test this in our application where master is not green yet (@daniellaniyo) is working on this.

rafaelfranca:
LGTM

Sirupsen:
bump @dylanahsmith

lib/identity_cache/configuration_dsl.rb
```diff
@@ -179,6 +179,10 @@ def disable_primary_cache_index
         self.primary_cache_index_enabled = false
       end
 
+      def should_use_cache?
```
dylanahsmith: I think you can remove this and just include the ShouldUseCache module in [IdentityCache.included](https://github.com/Shopify/identity_cache/blob/53b41864b78d822c29277db3e9faf0050768efb1/lib/identity_cache.rb#L57-L66)

lib/identity_cache/should_use_cache.rb
```diff
@@ -0,0 +1,13 @@
+module IdentityCache
+  module ShouldUseCache
+    extend ActiveSupport::Concern
+
+    included do
+      unless self.respond_to?(:should_use_cache?)
+        def self.should_use_cache?
```
dylanahsmith: Why can't you just define this as a normal module method with the `respond_to?` check?

Sirupsen: What do you mean by that? If I have `class ClassMethod` I can't run this. The purpose of this patch is to not override an existing `should_use_cache?`.

dylanahsmith: What method are you worrying about overriding?  The class that calls `include IdentityCache` just needs to define the override after that include.

Sirupsen: I'm concerned about this scenario:

```ruby
class AssociatedRecord < ActiveRecord::Base
  def self.should_use_cache?
    # some logic
  end
end

class Thing < ActiveRecord::Base
  include IdentityCache
  # this will override `AssociatedRecord#should_use_cache?`
  cache_has_many :associated_record 
end
```

dylanahsmith: That wouldn't be a problem if AssociatedRecord defined should_use_cache? directly like that, but I can see how that would be a problem if it was defined from a mixin.  However, I still don't like you you are dynamically defining should_use_cache? on the class itself here.

Instead, you could change

```ruby
child_class.send(:include, ShouldUseCache) unless child_class.include?(ShouldUseCache)
```

to

```
child_class.send(:include, ShouldUseCache) unless child_class.respond_to?(:should_use_cache?)
```

in add_parent_expiry_hook.

Sirupsen: Okay, you're out-rubbying me here Dylan, how does it make a difference whether `should_use_cache` is defined in a mixin in this case?

I'll change it to what you suggested. 👍

dylanahsmith: Including a module is like multiple inheritance, so it looks for the method in the class itself before looking for it in the included modules.  E.g.

```ruby
module Mixin
  def print
    puts 'mixin'
  end
end

class Thing
  def print
    puts 'thing'
  end
  include Mixin
end

Thing.new.print
```

prints `thing` because it found Thing#print before looking for Mixin#print.  However, what you were doing was basically just replacing Thing#print when `include Mixin` is called.

Sirupsen: Thanks for the explanation.

Sirupsen:
@dylanahsmith fixed the last `should_use_cache?` comments. 💨 ?

dylanahsmith:
:shipit:

PR #293, opened 2016-09-04
Support multiple databases and transactional tests in IdentityCache.should_use_cache?

eugeneius:
This implements both features described in https://github.com/Shopify/identity_cache/issues/274#issuecomment-227975692:
### Check for open transactions on all connection pools

IdentityCache disables cache reads inside transactions to try to prevent inconsistent cached data from being written back into the database.

When an application uses multiple databases, we need to check for open transactions on all of them; otherwise we could read stale data from a record stored in one database and write it into another database.
### Ignore the top level transaction in transactional tests

An unjoinable transaction forces any transactions nested inside it to behave as if they were at the top level: `after_commit` callbacks for records modified during the nested transaction will be executed when it commits, even though the updated data won't be visible to other connections until the outer transaction has also committed.

Rails creates unjoinable transactions in two places: in the sandbox console, and around transactional tests. In both of these cases, the user expects their code to behave as if the transaction wasn't there; with this change, IdentityCache will conform to that expectation by only disabling cache reads inside joinable transactions.

IdentityCache currently never reads from the cache during transactional tests, since there is always at least one transaction open. With this change transactional tests behave more like production code, which will help catch issues like updating a record that came from the cache.

### DIFF
--- a/lib/identity_cache.rb
+++ b/lib/identity_cache.rb
@@ -105,8 +105,9 @@ module IdentityCache
     end
 
     def should_use_cache? # :nodoc:
-      pool = ActiveRecord::Base.connection_pool
-      !pool.active_connection? || pool.connection.open_transactions == 0
+      ActiveRecord::Base.connection_handler.connection_pool_list.all? do |pool|
+        !pool.active_connection? || pool.connection.open_transactions == 0
+      end
     end
 
     # Cache retrieval and miss resolver primitive; given a key it will try to
--- a/lib/identity_cache.rb
+++ b/lib/identity_cache.rb
@@ -105,8 +105,26 @@ module IdentityCache
     end
 
     def should_use_cache? # :nodoc:
-      ActiveRecord::Base.connection_handler.connection_pool_list.all? do |pool|
-        !pool.active_connection? || pool.connection.open_transactions == 0
+      ActiveRecord::Base.connection_handler.connection_pool_list.none? do |pool|
+        pool.active_connection? &&
+          # Rails wraps each of your tests in a transaction, so that any changes
+          # made to the database during the test can be rolled back afterwards.
+          # These transactions are flagged as "unjoinable", which tries to make
+          # your application behave as if they weren't there. In particular:
+          #
+          #  - Opening another transaction during the test creates a savepoint,
+          #    which can be rolled back independently of the main transaction.
+          #  - When those nested transactions complete, any `after_commit`
+          #    callbacks for records modified during the transaction will run,
+          #    even though the changes haven't actually been committed yet.
+          #
+          # By ignoring unjoinable transactions, IdentityCache's behaviour
+          # during your test suite will more closely match production.
+          #
+          # When there are no open transactions, `current_transaction` returns a
+          # special `NullTransaction` object that is unjoinable, meaning we will
+          # use the cache.
+          pool.connection.current_transaction.joinable?
       end
     end
 
--- a/test/identity_cache_test.rb
+++ b/test/identity_cache_test.rb
@@ -9,6 +9,10 @@ class IdentityCacheTest < IdentityCache::TestCase
   class BadModel < BadModelBase
   end
 
+  class ModelWithConnection < ActiveRecord::Base
+    establish_connection ActiveRecord::Base.connection_config
+  end
+
   def test_identity_cache_raises_if_loaded_twice
     assert_raises(IdentityCache::AlreadyIncludedError) do
       BadModel.class_eval do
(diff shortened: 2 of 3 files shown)


camilo:
LGMT, but I won't merge quite yet @daniellaniyo, we need to test that this actually works on the shopi-side first.

dylanahsmith:
> we need to test that this actually works on the shopi-side first

We just override the method in Shopify, however, we should be able to use this logic there.  E.g. even if we override it, we should be able to use `super` in our override.

lib/identity_cache.rb
```diff
@@ -94,8 +94,9 @@ def should_fill_cache? # :nodoc:
     end
 
     def should_use_cache? # :nodoc:
-      pool = ActiveRecord::Base.connection_pool
-      !pool.active_connection? || pool.connection.open_transactions == 0
+      ActiveRecord::Base.connection_handler.connection_pool_list.none? do |pool|
+        pool.active_connection? && pool.connection.current_transaction.joinable?
```
dylanahsmith: It seems like this would normally work, because non-joinable transactions is an obscure feature that internally is only used for the console sandbox and transactional fixtures.  However, that same obscurity of the feature makes this code hard to understand, along with relying on current_transaction to return NullTransaction.

Could you comment the code so that it can be read without also reading the active record internals?

test/identity_cache_test.rb
```diff
@@ -23,4 +27,16 @@ def test_should_use_cache_in_transaction
...
+    ModelWithConnection.transaction do
+      assert_equal false, IdentityCache.should_use_cache?
+    end
+  end
+
+  def test_should_use_cache_inside_unjoinable_transaction
+    ActiveRecord::Base.transaction(joinable: false) do
+      assert_equal true, IdentityCache.should_use_cache?
+    end
+  end
```
dylanahsmith: Could you add a test case that uses `self.use_transactional_fixtures = false` so that you can test that should_use_cache? works without a transaction (i.e. the NullTransaction)

lib/identity_cache.rb
camilo: @rafaelfranca @sgrif wdyt about relying on this feature ?

eugeneius: I added a (perhaps too long) comment.

The current implementation uses `TransactionManager#open_transactions`, which is also not part of the public Rails API - I do agree that `#joinable?` is a lot harder to understand, though.

test/identity_cache_test.rb
eugeneius: `IdentityCache::TestCase` doesn't include `ActiveRecord::TestFixtures`, so `use_transactional_fixtures` isn't available and none of the tests are wrapped in transactions.

My intention here was that this new test would cover the transactional fixtures case by manually opening an unjoinable transaction, and [this existing test](https://github.com/Shopify/identity_cache/blob/v0.3.2/test/identity_cache_test.rb#L17-L19) would cover the case where no transactions are open and a `NullTransaction` is returned.

However the intention of the tests is clearer when `use_transactional_fixtures` is used, so I've changed the tests to use different test case classes that include `ActiveRecord::TestFixtures` and set `use_transactional_fixtures`.

test/identity_cache_test.rb
```diff
@@ -14,6 +18,42 @@ def test_identity_cache_raises_if_loaded_twice
...
+end
+
+class IdentityCacheWithTransactionalFixturesTest < ActiveSupport::TestCase
+  include ActiveRecord::TestFixtures
+  if respond_to?(:use_transactional_tests)
+    self.use_transactional_tests = false
+  else
+    self.use_transactional_fixtures = false
```
dylanahsmith: Shouldn't these be set to true for this test case class?

eugeneius: Yes 🙈 updated!

eugeneius: Rather, updated in `IdentityCacheWithTransactionalFixturesTest` above ☝️

dylanahsmith:
LGTM

lib/identity_cache.rb
rafaelfranca: This make sense to me.

eugeneius:
👋 @camilo - did you get a chance to try this out in the Shopify codebase?

daniellaniyo:
@eugeneius This is on hold because some external tests were failing on the shopify side. Will let you know as soon as there are further updates.

eugeneius:
I just rebased this, fixed the conflicts, and signed the CLA (which didn't exist when I opened it 😄).

Does this patch still cause test failures in Shopify's codebase? If so, maybe your version of the monkey patch could be upstreamed instead?

dylanahsmith:
For reference, rails itself uses [`run_commit_callbacks = !current_transaction.joinable?`](https://github.com/rails/rails/blob/0ecf163cc8851e46b680306d328c194a3b38e2c3/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb#L226) to determine whether commit callbacks will be called, which matches what is being done in this PR.

Shopify is currently replacing the `should_use_cache?` implementation, so this shouldn't affect it.  Shopify should leverage rails 6.1 sharding support, but we shouldn't block this PR because of that technical debt.  It has been sitting way too long without a serious concern.

We can always fix forward if we find a problem with this.  However, this clearly appears better than the code it replaces.

PR #294, opened 2016-09-05
Fix fetching already loaded belongs_to association

eugeneius:
This path was missed when `load_and_readonlyify` was refactored into `readonly_copy` in https://github.com/Shopify/identity_cache/pull/288, and there was no test covering the behaviour.


camilo:
@dylanahsmith you might want to review this one since you have more context.

dylanahsmith:
Thanks.  Sorry for missing that

PR #296, opened 2016-10-01
Move delete failure detection to the cache fetcher.

### LINKED ISSUE #211: Deletes failing and cache not expiring on one particular model/association
I've got a weird issue with deletes failing when expiring. We use identity_cache in lots of places and it mostly works fine, except on this one particular model's association we're having problems. Here's an example from the console:

```shell
[2] (console)> app.fetch_forward_mail_setup_attempts.count
   (0.4ms)  SELECT `forward_mail_setup_attempts`.id FROM `forward_mail_setup_attempts`  WHERE `forward_mail_setup_attempts`.`app_id` = 41
Cache read_multi: ["IDC:5:blob:ForwardMailSetupAttempt:9118014673441475248:73", "IDC:5:blob:ForwardMailSetupAttempt:9118014673441475248:74", "IDC:5:blob:ForwardMailSetupAttempt:9118014673441475248:75", "IDC:5:blob:ForwardMailSetupAttempt:9118014673441475248:76", "IDC:5:blob:ForwardMailSetupAttempt:9118014673441475248:77"]
Dalli::Server#connect localhost:11211
[IdentityCache] (backend) cache hit for IDC:5:blob:ForwardMailSetupAttempt:9118014673441475248:73 (multi)
[IdentityCache] (backend) cache hit for IDC:5:blob:ForwardMailSetupAttempt:9118014673441475248:74 (multi)
[IdentityCache] (backend) cache hit for IDC:5:blob:ForwardMailSetupAttempt:9118014673441475248:75 (multi)
[IdentityCache] (backend) cache hit for IDC:5:blob:ForwardMailSetupAttempt:9118014673441475248:76 (multi)
[IdentityCache] (backend) cache hit for IDC:5:blob:ForwardMailSetupAttempt:9118014673441475248:77 (multi)
=> 5
[3] (console)> app.forward_mail_setup_attempts.create!(
[3] (console)*   forwarding_email_address: "b@b.com",  
[3] (console)*   started_at: Time.now,  
[3] (console)*   expired_at: Time.now + 1.day  
[3] (console)* )  
   (0.4ms)  BEGIN
  SQL (6.6ms)  INSERT INTO `forward_mail_setup_attempts` (`app_id`, `created_at`, `expired_at`, `forwarding_email_address`, `started_at`, `updated_at`) VALUES (41, '2015-05-29 18:08:09', '2015-05-30 18:08:09', 'b@b.com', '2015-05-29 18:08:09', '2015-05-29 18:08:09')
   (11.3ms)  COMMIT
[IdentityCache] expiring=App expiring_id=41 expiring_last_updated_at=2015-05-28 23:25:03 UTC
Cache delete: IDC:5:blob:App:18240036282024875059:41
[IdentityCache] delete failed for IDC:5:blob:App:18240036282024875059:41
[IdentityCache] expiring=ForwardMailSetupAttempt expiring_id=78 expiring_last_updated_at=
Cache delete: IDC:5:blob:ForwardMailSetupAttempt:9118014673441475248:78
[IdentityCache] delete failed for IDC:5:blob:ForwardMailSetupAttempt:9118014673441475248:78
=> #<ForwardMailSetupAttempt id: 78, app_id: 41, forwarding_email_address: "b@b.com", provider: nil, started_at: "2015-05-29 18:08:09", expired_at: "2015-05-30 18:08:09", created_at: "2015-05-29 18:08:09", updated_at: "2015-05-29 18:08:09">
[4] (console)> app.fetch_forward_mail_setup_attempts.count
=> 5
```

When I create a new record the deletes fail and the cache doesn't expire. I tailed memcached logs and see messages like `Writing an error: Not found`.

Any idea what might be going wrong?

dylanahsmith:
@daniellaniyo, @Sirupsen please review

Fixes #211

## Problem

IdentityCache::MemoizedCacheProxy#delete was assuming the delete failed if cache_fetcher.delete returned falsey.  However, that falsey value could either mean the memcached request failed or the key wasn't present, and different cache stores aren't consistent about whether `nil` or `false` means as a return value for their delete method as I detailed in https://github.com/Shopify/identity_cache/issues/211#issuecomment-137626605.

This was confusing because we were logging something like `[IdentityCache] delete failed for IDC:5:blob:App:18240036282024875059:41` which sounds like a probably with expiring the cache, but it could just mean that the value wasn't cached yet.

## Solution

If IdentityCache::MemoizedCacheProxy#cache_fetcher is an IdentityCache::CacheFetcher, then it writes a IdentityCache::DELETED value instead of using a delete memcached request.  In that case the return value isn't as ambiguous, it just returns falsey if the request failed.  So, I was able to move the failure detection into IdentityCache::CacheFetcher#delete where we can rely on it and put a generic `[IdentityCache] delete for #{key}` message in IdentityCache::FallbackFetcher#delete.

### DIFF
--- a/lib/identity_cache/cache_fetcher.rb
+++ b/lib/identity_cache/cache_fetcher.rb
@@ -11,7 +11,9 @@ module IdentityCache
     end
 
     def delete(key)
-      @cache_backend.write(key, IdentityCache::DELETED, :expires_in => IdentityCache::DELETED_TTL.seconds)
+      result = @cache_backend.write(key, IdentityCache::DELETED, expires_in: IdentityCache::DELETED_TTL.seconds)
+      IdentityCache.logger.debug { "[IdentityCache] delete #{ result ? 'recorded' : 'failed' } for #{key}" }
+      result
     end
 
     def clear
--- a/lib/identity_cache/fallback_fetcher.rb
+++ b/lib/identity_cache/fallback_fetcher.rb
@@ -11,7 +11,9 @@ module IdentityCache
     end
 
     def delete(key)
-      @cache_backend.delete(key)
+      result = @cache_backend.delete(key)
+      IdentityCache.logger.debug { "[IdentityCache] delete for #{key}" }
+      result
     end
 
     def clear
--- a/lib/identity_cache/memoized_cache_proxy.rb
+++ b/lib/identity_cache/memoized_cache_proxy.rb
@@ -36,9 +36,7 @@ module IdentityCache
 
     def delete(key)
       memoized_key_values.delete(key) if memoizing?
-      result = @cache_fetcher.delete(key)
-      IdentityCache.logger.debug { "[IdentityCache] delete #{ result ? 'recorded'  : 'failed'  } for #{key}" }
-      result
+      @cache_fetcher.delete(key)
     end
 
     def fetch(key)


lib/identity_cache/cache_fetcher.rb
```diff
@@ -11,7 +11,9 @@ def write(key, value)
     end
 
     def delete(key)
-      @cache_backend.write(key, IdentityCache::DELETED, :expires_in => IdentityCache::DELETED_TTL.seconds)
+      result = @cache_backend.write(key, IdentityCache::DELETED, expires_in: IdentityCache::DELETED_TTL.seconds)
+      IdentityCache.logger.debug { "[IdentityCache] delete #{ result ? 'recorded' : 'failed' } for #{key}" }
```
Sirupsen: The reason you're using blocks here is to avoid allocating the logging string unless you're at the debugging level?

lib/identity_cache/fallback_fetcher.rb
```diff
@@ -11,7 +11,9 @@ def write(key, value)
     end
 
     def delete(key)
-      @cache_backend.delete(key)
+      result = @cache_backend.delete(key)
```
Sirupsen: Why is it OK to do this in the fallback fetcher? What is the fallback fetcher?

lib/identity_cache/cache_fetcher.rb
dylanahsmith: I was just doing it the way we were doing it before, but I think that is probably the reason.

lib/identity_cache/fallback_fetcher.rb
dylanahsmith: The fallback fetcher is just for compatibility for cache backends that don't support CAS operations.  [memcached_store](https://rubygems.org/gems/memcached_store) is the only active support cache store that I know of that supports CAS.

The difference is that now it will just unconditionally log `[IdentityCache] delete for #{key}` with debug logging and won't attempt to indicate whether it failed or not.

dylanahsmith:
Sure, I'll add a test for this.  Also, I think we should probably use a higher log level for logging that the delete failed, since normally we wouldn't log at the debug level in production, yet this could be useful to investigating cache corruption.

lib/identity_cache/cache_fetcher.rb
camilo: That is the reason.

camilo:
So, TL;DR the one reason to do this is to make the non-CAS capable adaptors logging less noisy? Or is most of that logging just useless now?

This LGTM in isolation but I wonder if we should make CAS a requirement it has improved a bunch of cache desynchronization things for us 

cc @fbogsany

PR #297, opened 2016-10-13
Avoid querying when no ids are passed to fetch_multi

pushrax:
When `should_use_cache?` is false, and no ids are passed to `fetch_multi` on the model, we unnecessarily send an impossible query. ActiveRecord generates something like `SELECT ... WHERE 1=0`. May as well avoid the roundtrip, like `IdentityCache.fetch_multi` does [here](https://github.com/Shopify/identity_cache/blob/e5e0edf653d1295964134a732a854231357be47a/lib/identity_cache.rb#L131).

@dylanahsmith @Sirupsen @fw42

### DIFF
--- a/test/fetch_multi_test.rb
+++ b/test/fetch_multi_test.rb
@@ -265,6 +265,16 @@ class FetchMultiTest < IdentityCache::TestCase
     end
   end
 
+  def test_fetch_multi_with_no_keys_does_not_query_when_cache_is_disabled
+    Item.stubs(:should_use_cache?).returns(false)
+
+    assert_queries(0) do
+      assert_memcache_operations(0) do
+        Item.fetch_multi
+      end
+    end
+  end
+
   private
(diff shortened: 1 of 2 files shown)


test/fetch_multi_test.rb
```diff
@@ -265,6 +265,12 @@ def test_fetch_multi_with_open_transactions_returns_non_readonly_records
     end
   end
 
+  def test_fetch_multi_with_no_keys_does_not_query_when_cache_is_disabled
+    Item.stubs(:should_use_cache?).returns(false)
+    Item.expects(:where).never
```
dylanahsmith: never expectations are extra brittle.  I would prefer if `assert_queries(0)` and `assert_memcache_operations(0)` were used

pushrax: ah cool 👍

PR #302, opened 2016-12-19
Lazy loading issues

casperisfine:
Fix: https://github.com/Shopify/identity_cache/issues/300

WIP

PR #303, opened 2016-12-25
refactor: Include modules into IdentityCache instead into models

dylanahsmith:
## Problem

When looking at https://github.com/Shopify/identity_cache/pull/302 I was noticing how much of an unnecessary mess our module includes are in IdentityCache.  We were doing a lot of dynamic module inclusions in included blocks/methods to handle module dependancies when this is something that ActiveSupport::Concern can handle for us.

I would like this refactored to make the code more readable.

## Solution

Just include the module dependancies directly in the IdentityCache module rather than doing it dynamically.  We don't need to check if a module is already included because ruby won't insert a module into the ancestors for a class multiple times and ActiveSupport::Concern will only call the included block for module dependancies once even if the module is included multiple times.

I also tried to make the purpose for the AlreadyIncluded error more clear by modifying the test to use an actual problematic example where IdentityCache is included into the base and derived model.  For the same reason I also co-located the code to setup cached_model along with the AlreadyIncluded check.

### DIFF
--- a/lib/identity_cache.rb
+++ b/lib/identity_cache.rb
@@ -69,6 +69,7 @@ module IdentityCache
       raise AlreadyIncludedError if base.respond_to?(:cached_model)
       base.class_attribute :cached_model
       base.cached_model = base
+      super
     end
 
     # Sets the cache adaptor IdentityCache will be using
--- a/lib/identity_cache.rb
+++ b/lib/identity_cache.rb
@@ -7,8 +7,8 @@ require 'identity_cache/memoized_cache_proxy'
 require 'identity_cache/belongs_to_caching'
 require 'identity_cache/cache_key_generation'
 require 'identity_cache/configuration_dsl'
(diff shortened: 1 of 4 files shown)


lib/identity_cache.rb
```diff
@@ -56,15 +66,9 @@ class << self
     self.fetch_read_only_records = version >= Gem::Version.new("0.5")
 
     def included(base) #:nodoc:
```
rafaelfranca: If we are using `concern` we should call `super` here or use `included` block instead of defining the method.

dylanahsmith: In this case I don't want to use the included block since I still want to prevent the module from being included into a base and derived class.  ActiveSupport::Concern in this module is being used to manage the dependant modules. I've added `super` to this included method override though.

rafaelfranca: :+1:

PR #308, opened 2017-02-02
Oops fix syntax error in performance code.

simon-shopify:
I think [it's a copy paste error](https://github.com/Shopify/identity_cache/commit/6c44ed4ce2f7e5e9aa39d703f20ff15e44f58248#diff-be43b92f25d15301aa4dd20dc137bf38R18) and meant to be an `if`. Anyways, it fixes `dev benchmark-cpu`.

r: @casperisfine @tjoyal

### DIFF
--- a/performance/cache_runner.rb
+++ b/performance/cache_runner.rb
@@ -15,7 +15,7 @@ require File.dirname(__FILE__) + '/../test/helpers/database_connection'
 IdentityCache.logger = Logger.new(nil)
 IdentityCache.cache_backend = ActiveSupport::Cache::MemcachedStore.new("localhost:#{$memcached_port}", :support_cas => true)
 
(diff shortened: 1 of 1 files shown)


performance/cache_runner.rb
```diff
@@ -15,7 +15,7 @@
 IdentityCache.logger = Logger.new(nil)
 IdentityCache.cache_backend = ActiveSupport::Cache::MemcachedStore.new("localhost:#{$memcached_port}", :support_cas => true)
 
-ActiveRecord.gem_version < Gem::Version.new('5') && ActiveRecord::Base.respond_to?(:raise_in_transactional_callbacks=)
+if ActiveRecord.gem_version < Gem::Version.new('5') && ActiveRecord::Base.respond_to?(:raise_in_transactional_callbacks=)
```
casperisfine: That will always be false now. Might as well delete it all.

casperisfine: 🤦‍♂️ nvm, identity_cache can stillb e used by 4.2 code bases.

PR #309, opened 2017-02-02
Prepare for release 0.4.1.

simon-shopify:
There were a bunch of changes sitting on `master`. Let's release them?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # IdentityCache changelog
 
+#### 0.4.1
+
+- Deprecated embedded associations on models that don't use IDC (#305)
+- Remove a respond_to? check that hides mistakes in includes hash (#307)
+- Drop ruby 2.1 support (#301)
+- Avoid querying when no ids are passed to fetch_multi (#297)
+- Fix fetching already loaded belongs_to association (#294)
+- Move `should_use_cache?` calls to the model-level (#291)
+- Clone instead of dup record when readonlyifying fetched records (#292)
+- Consistently store the array for cached has many associations (#288)
+
 #### 0.4.0
 
 - Return an array from fetched association to prevent chaining. Up to now, a relation was returned by default. (#287)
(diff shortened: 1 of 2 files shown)


lib/identity_cache.rb
```diff
@@ -59,13 +59,13 @@ class << self
...
     mattr_accessor :never_set_inverse_association
-    self.never_set_inverse_association = version >= Gem::Version.new("0.5")
+    self.never_set_inverse_association = version >= Gem::Version.new("0.6")
 
     # Fetched records are not read-only and this could sometimes prevent IDC from
     # reflecting what's truly in the database when fetch_read_only_records is false.
     # When set to true, it will only return read-only records when cache is used.
     mattr_accessor :fetch_read_only_records
-    self.fetch_read_only_records = version >= Gem::Version.new("0.5")
+    self.fetch_read_only_records = version >= Gem::Version.new("0.6")
```
casperisfine: That's super weird. Who wrote that?

dylanahsmith: these are breaking changes that we want to make for 0.5.  Don't just bump the number.  They are configurations so we can always use the old behaviour in Shopify until we can actually turn these on.

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # IdentityCache changelog
 
+#### 0.5.0
+
+- Deprecated embedded associations on models that don't use IDC (#305)
+- Remove a respond_to? check that hides mistakes in includes hash (#307)
+- refactor: Include modules into IdentityCache instead into models (#303)
```
dylanahsmith: Refactoring shouldn't matter to the user's of the library.  Only include changes that affect the users of the library

dylanahsmith:
Are there any changes that necessitate a 0.5 release rather than a 0.4.1 release?

lib/identity_cache.rb
casperisfine: Right, but then now that we are shipping 0.6? we can simply resolve that to `true`

simon-shopify:
> Are there any changes that necessitate a 0.5 release rather than a 0.4.1 release?

Yeah I don't think there are any breaking changes, I can make it 0.4.1.

PR #310, opened 2017-02-02
Update tests for Railgun.

simon-shopify:
This is needed because we use real databases to snapshot IDC payload to test for format stability.

Please review @Shopify/webscalenext.


simon-shopify:
Oops looks like I need to be smarter because it breaks on CI.

simon-shopify:
There are some weird test failures I think I'm doing something wrong.

PR #311, opened 2017-02-02
Store the class name instead of the class itself.

simon-shopify:
This PR changes how records are store so we store the name of the class instead of the class itself. The goal of this is to make it easier to work with cache backend that do not support arbitrary Ruby objects.

This change is backward compatible: it still handles class on deserialization.


simon-shopify:
I'll extract the `pry` stuff in another PR.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # IdentityCache changelog
 
+#### 0.4.2
```
tjoyal: I think this should read:
```
#### Unreleased
```

lib/identity_cache/query_api.rb
```diff
@@ -122,7 +122,12 @@ def check_association_scope(association_name)
 
       def record_from_coder(coder) #:nodoc:
         if coder
-          klass = coder[:class]
+          klass =
+            if coder[:class].is_a?(String)
+              coder[:class].constantize
+            else
+              coder[:class]
```
tjoyal: What would be a good way to mark(or remember) this as deprecated to make sure we remove it in a future update?

dylanahsmith: Changing how we write data to memcache means we would have to bump the CACHE_VERSION.  We use that as part of the cache key, so once that is bumped we wouldn't have to support the deprecated code path.  It would also mean the cache would effectively get flushed as a result.

CHANGELOG.md
dylanahsmith: Maybe we should have a larger version bump if the cache format is changing.

simon-shopify:
> Changing how we write data to memcache means we would have to bump the CACHE_VERSION. We use that as part of the cache key, so once that is bumped we wouldn't have to support the deprecated code path. It would also mean the cache would effectively get flushed as a result.

I just wonder what would happen on the following scenario:
* Upgrade with the new `CACHE_VERSION`
* Records start to get invalidated according to the new version.
* We rollback.
* We have a bunch of stale record sitting in IDC.

In that case we could de two things
1. Just flush the cache manually to make sure we don't get any stale entries. Can Shopify handle a cache flush?
2. Roll out the backwards compatible code to read the entries, then roll out the new code that writes the entries?

I think you are right and bumping the cache version + flushing the cache is the easiest thing, but I just want to make sure we can handle a cache flush.

dylanahsmith:
Bumping the cache version is safer, since we don't end up with exceptions from old code trying to read the new cache keys.  But you're right, that in the case of a rollback we would want to flush the cache.

It would be nice if we made it easier to manage this type of short transition.  For this use case it would be ideal if we instead stored the cache key format as part of the value instead of the key so that we could treat it as a cache miss if the format doesn't match which would automatically handle the case where we rollback.  This would also make it possible to support reading data for the next cache version to allow for the possibility of a seamless transition between cache versions.

On the other hand, longer rollouts have a different requirement of being able to cache data in separate cache keys for different cache formats when upgrading ruby or rails versions, and have both old and new cache keys for data get invalidated.  So far we have had to monkey patch IDC to support this scenario.

For now I think we can still get away with just bumping the cache version and flushing IDC if a rollback is needed, but check with prod-eng to make they are cool with IDC being flushed.

dylanahsmith:
Actually, I think IDC should now be working across rails versions.  If you are working on moving away from using Marshal in the cache backend, then we won't have to worry about changes between ruby versions as well.

simon-shopify:
* Updated to bump `CACHE_VERSION`.

simon-shopify:
Is it best if I bump the version in this PR or if I do a second PR after? What do you think of bumping the minor to `0.5.0`?

dylanahsmith:
Sure, you could bump the minor version in this PR.

simon-shopify:
> Sure, you could bump the minor version in this PR.

Ah I just remember there are [these conditions on the version number](https://github.com/Shopify/identity_cache/blob/e03e11e430a0d6a072cc7c5215d590e6764c8409/lib/identity_cache.rb#L62). I'll do a separate PR where I clean it up instead.

simon-shopify:
@dylanahsmith do you have any other concern before I merge this?

PR #312, opened 2017-02-02
Remove Postgres from Homebrew.

simon-shopify:
It is no longer necessary when using dev since we now have it in Railgun.

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -1,8 +1,6 @@
 name: identity-cache
 
 up:
-  - homebrew:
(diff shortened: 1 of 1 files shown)


rafaelfranca:
Don't we need it for the client?

simon-shopify:
Oops, [you are right](https://bitbucket.org/ged/ruby-pg/wiki/Home). For some reason I thought it was bundled with the gem. I'll revert it.

PR #313, opened 2017-02-03
Add Pry in development.

simon-shopify:
I ended up adding and removing it a few time, what do you think if I add it for real?


rafaelfranca:
Pry, debugger, byebug... Each developer use a different debugger and maybe
choosing one as a dependency of the project is not a good idea. If I'd
chose one it would be byebug because it is officially supported by Ruby
Core team.

simon-shopify:
> Pry, debugger, byebug... Each developer use a different debugger and maybe
> choosing one as a dependency of the project is not a good idea. If I'd
> chose one it would be byebug because it is officially supported by Ruby
> Core team.

Yeah I think you're right.

PR #314, opened 2017-02-06
Revert "Remove Postgres from Homebrew."

simon-shopify:
Reverts Shopify/identity_cache#312


simon-shopify:
cc: @rafaelfranca @dylanahsmith

PR #317, opened 2017-02-13
Fix infinite caching in the case of self associated models

dzhangd:
When we tried to cache a model with itself and had `embed: true`, identity cache would attempt to cache the objects in an infinite loop and caused a stack overflow. This PR aims to fix that by adding checks for that case.

### DIFF
--- a/identity_cache.gemspec
+++ b/identity_cache.gemspec
@@ -26,7 +26,6 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency('mocha', '0.14.0')
   gem.add_development_dependency('spy')
   gem.add_development_dependency('minitest', '>= 2.11.0')
-  gem.add_development_dependency('pry-byebug')
 
   if RUBY_PLATFORM == 'java'
     raise NotImplementedError
--- a/identity_cache.gemspec
+++ b/identity_cache.gemspec
@@ -26,6 +26,7 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency('mocha', '0.14.0')
   gem.add_development_dependency('spy')
   gem.add_development_dependency('minitest', '>= 2.11.0')
+  gem.add_development_dependency('pry-byebug')
 
   if RUBY_PLATFORM == 'java'
     raise NotImplementedError
--- a/lib/identity_cache/cache_key_generation.rb
+++ b/lib/identity_cache/cache_key_generation.rb
@@ -7,14 +7,17 @@ module IdentityCache
       columns.sort_by(&:name).map{|c| "#{c.name}:#{c.type}"}.join(',')
     end
 
-    def self.denormalized_schema_hash(klass)
+    def self.denormalized_schema_hash(klass, seen_assocs)
       schema_string = schema_to_string(klass.columns)
       klass.send(:all_cached_associations).sort.each do |name, options|
         klass.send(:check_association_scope, name)
         ParentModelExpiration.check_association(options) if options[:embed]
         case options[:embed]
         when true
-          schema_string << ",#{name}:(#{denormalized_schema_hash(options[:association_reflection].klass)})"
+          unless seen_assocs.add?(name)
+              return IdentityCache.memcache_hash(schema_to_string(klass.columns))
+          end
+          schema_string << ",#{name}:(#{denormalized_schema_hash(options[:association_reflection].klass, seen_assocs)})"
         when :ids
           schema_string << ",#{name}:ids"
         end
@@ -28,7 +31,7 @@ module IdentityCache
       end
 
       def rails_cache_key_prefix
-        @rails_cache_key_prefix ||= IdentityCache::CacheKeyGeneration.denormalized_schema_hash(self)
+        @rails_cache_key_prefix ||= IdentityCache::CacheKeyGeneration.denormalized_schema_hash(self, Set.new)
       end
 
       def prefixed_rails_cache_key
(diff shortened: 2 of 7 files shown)


identity_cache.gemspec
```diff
@@ -26,6 +26,7 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency('mocha', '0.14.0')
   gem.add_development_dependency('spy')
   gem.add_development_dependency('minitest', '>= 2.11.0')
+  gem.add_development_dependency('pry-byebug')
```
simon-shopify: You need to remove this.

lib/identity_cache/cache_key_generation.rb
```diff
@@ -7,14 +7,17 @@ def self.schema_to_string(columns)
...
+    def self.denormalized_schema_hash(klass, seen_assocs = Set.new)
       schema_string = schema_to_string(klass.columns)
       klass.send(:all_cached_associations).sort.each do |name, options|
         klass.send(:check_association_scope, name)
         ParentModelExpiration.check_association(options) if options[:embed]
         case options[:embed]
         when true
-          schema_string << ",#{name}:(#{denormalized_schema_hash(options[:association_reflection].klass)})"
+          unless seen_assocs.add?(name)
+              return IdentityCache.memcache_hash(schema_to_string(klass.columns))
```
simon-shopify: That looks wrong. It will skip all subsequent associations. Perhaps you mean only `next`?

test/helpers/database_connection.rb
```diff
@@ -50,6 +50,8 @@ def self.create_tables
     :not_cached_records            => [[:string, :name], [:integer, :item_id], [:timestamps, null: true]],
     :items                         => [[:integer, :item_id], [:string, :title], [:timestamps, null: true]],
     :items2                        => [[:integer, :item_id], [:string, :title], [:timestamps, null: true]],
+    self_items:                       [[:string, :title], [:integer, :parent_item_id], [:integer, :self_item_two_id], [:timestamps, null: true]],
+    self_item_twos:                   [[:string, :title], [:integer, :self_item_id], [:timestamps, null: true]],
```
simon-shopify: Either use hash rockets here or change all of them to the new syntax.

dylanahsmith:
> When we tried to cache a model with itself and had `embed: true`, identity cache would attempt to cache the objects in an infinite loop and caused a stack overflow.

That is because it embeds recursively.

> This PR aims to fix that by adding checks for that case.

In this case it should just raise an exception and recommend using an id embedded cached association instead.

dzhangd:
@dylanahsmith @simon-shopify, @gabriel-letarte and I had a discussion and we decided that in the use case that we're thinking of, caching all the models as a blob instead of doing a `fetch_multi` with IDs makes more sense. Maybe we should sync up and discuss this some more?

identity_cache.gemspec
dzhangd: Wouldn't it be useful for future debugging? It's only a development dependency.

simon-shopify:
> all the models as a blob instead of doing a fetch_multi with IDs makes more sense.

I think you mentioned that reason, but I can't remember what it is. Can you write it here?

> That is because it embeds recursively.

These models are nodes in a tree, so can never form cycles.

identity_cache.gemspec
simon-shopify: https://github.com/Shopify/identity_cache/pull/313

lib/identity_cache/cache_key_generation.rb
dzhangd: No, that's the intended behaviour. When we see a model that we've already encountered, that model's associations have already put into the cache higher up in the call stack, so there's no need to put them in the cache key again.

simon-shopify: I meant the other classes in `klass.send(:all_cached_associations).sort`.

dzhangd: Yeah, because for `seen_assocs.add?` to falsey, the current class must have been added to `seen_assocs` sometime before in a previous call up the call stack. In that call, `unless seen_assocs.add?(name)` wouldn't pass and we would continue execution onto the other elements in `klass.send(:all_cached_associations).sort`.

In other words, there's no point of continuing execution if we've seen an association because by that point, all that model's associations have already been accounted for.

lib/identity_cache/cache_key_generation.rb
```diff
@@ -7,14 +7,17 @@ def self.schema_to_string(columns)
...
-    def self.denormalized_schema_hash(klass)
+    def self.denormalized_schema_hash(klass, seen_assocs = Set.new)
       schema_string = schema_to_string(klass.columns)
       klass.send(:all_cached_associations).sort.each do |name, options|
         klass.send(:check_association_scope, name)
         ParentModelExpiration.check_association(options) if options[:embed]
         case options[:embed]
         when true
-          schema_string << ",#{name}:(#{denormalized_schema_hash(options[:association_reflection].klass)})"
+          unless seen_assocs.add?(name)
```
dylanahsmith: You would need to use the association klass as well as the name.  Otherwise, this might have false positives where it finds two associations on different classes with the same name and assume the association has already been added when it hasn't been.

lib/identity_cache/cache_key_generation.rb
dylanahsmith: It sounds like you should be storing a set of classes in the set rather than associations.

lib/identity_cache/query_api.rb
```diff
@@ -277,11 +277,12 @@ def embedded_associations
...
 
-      def cache_fetch_includes
+      def cache_fetch_includes(seen_assocs = Set.new)
         associations_for_identity_cache = recursively_embedded_associations.map do |child_association, options|
           child_class = reflect_on_association(child_association).try(:klass)
-
-          child_includes = child_class.send(:cache_fetch_includes)
+          if seen_assocs.add?(child_association)
+            child_includes = child_class.send(:cache_fetch_includes, seen_assocs)
+          end
```
dylanahsmith: Active record includes hashes won't recursively fetch associations, so would need a lot large refactor if we chose so support it

PR #318, opened 2017-02-13
Alternate cache namespace for smooth rollouts.

simon-shopify:
Every time we roll out a breaking change to IDC we have to monkey patch it to delete both versions when a value is expired. No more... well yes, but it's now much easier. You have to set `IdentityCache. alternate_cache_namespace` and you are set!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # IdentityCache changelog
 
+#### Unreleased
+
+- Support alternate cache key for smoother upgrades
+
 #### 0.5.1
 
 - Fix bug in prefetch_associations for cache_has_one associations that may be nil
--- a/lib/identity_cache.rb
+++ b/lib/identity_cache.rb
@@ -49,6 +49,11 @@ module IdentityCache
     mattr_accessor :cache_namespace
     self.cache_namespace = "IDC:#{CACHE_VERSION}:".freeze
 
+    # If non-nil, IDC will expire entries on that namespace as well. Use this to
+    # progressively roll out a new cache format.
+    mattr_accessor :alternate_cache_namespace
+    self.alternate_cache_namespace = nil
+
     # Inverse active record associations are set when loading embedded
     # cache_has_many associations from the cache when never_set_inverse_association
     # is false. When set to true, it will only set the inverse cached association.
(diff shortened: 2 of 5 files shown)


dylanahsmith:
> Every time we roll out a breaking change to IDC we have to monkey patch it to delete both versions when a value is expired.

I thought we were able to roll out the last breaking change by just bumping the cache version.

Is this needed to try a slow rollout of a change to the cache format (e.g. marshal -> messagepack)?

Also, this seems like a big diff for something that only required the following small monkey patch:

```ruby
module IdentityCache
  module SlowRolloutSupport
    def legacy_namespace
      @legacy_namespace ||= "#{Shopify::Config.marshal_key_prefix}:S:"
    end

    def delete(key)
      legacy_key = key.sub(/\A#{IdentityCache::CacheKeyGeneration::DEFAULT_NAMESPACE}/o, "IDC:5:")
      cache_backend.write(legacy_key, IdentityCache::DELETED, expires_in: IdentityCache::DELETED_TTL.seconds, namespace: legacy_namespace)
      IdentityCache.logger.debug { "[IdentityCache] delete legacy key: '#{legacy_namespace}:#{legacy_key}'" }
      super
    end
  end
  CacheFetcher.prepend(SlowRolloutSupport)
end
```

Actually, it looks like maybe this could just be implemented by wrapping the cache backend rather than monkey patching.

casperisfine:
> I thought we were able to roll out the last breaking change by just bumping the cache version.

Not really. If you simply bump the cache version, you incur many risk of during the rollout, and make rollback very dangerous.

Imagine the following scenario during the rollout:

  - Worker 1 is running the new version. It fetches a record and cache it in the new namespace.
  - Worker 2 is still running the old version. It handle a write, and as such invalidate the key in the old namespace.
  - Until that record is modified again, IDC will serve an outdated cache for that payload.

So for a proper smooth rollout you need:
  - Ship code that delete the future new namespace in parallel.
  - Ship code with the new cache version, and configured to delete the old namespace in parallel.
  - Once you are sure you won't have to rollback, you can stop deleting the old namespace.

> Actually, it looks like maybe this could just be implemented by wrapping the cache backend rather than monkey patching.

I totally agree. A simple decorator would do just fine.

dylanahsmith:
If the use case is a quick rollout, then the best solution would be to store the cache version in the cache value. The we could safely bump the version in one step.

casperisfine:
> If the use case is a quick rollout, then the best solution would be to store the cache version in the cache value. The we could safely bump the version in one step.

It would indeed be easier to rollout. I'd be 👍 with such solution.

simon-shopify:
Yeah I expected it to be simpler, but the code got out of hand 😅 . Since I'm changing the codec under IDC, I'll bake the codec version in the payload for now and we can revisit having a version in the IDC data later.

PR #319, opened 2017-02-17
Remote deprecated never_set_inverse_association option.

simon-shopify:
I removed the deprecated option from Shopify, so I figured I could clean it up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # IdentityCache changelog
 
+#### 0.6.0 Unreleased
+
+- Remove deprecated `never_set_inverse_association` option (#319)
+
 #### 0.5.1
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # IdentityCache changelog
 
+#### Unreleased
```
dylanahsmith: Maybe this should be `#### 0.6.0 Unreleased` to make it clear that the next release will have breaking changes.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # IdentityCache changelog
 
+#### 0.6.0 Unreleased
```
tjoyal: Since it was already the default and deprecated, it might only require a minor? 

I really don't care about administration :+1:

PR #322, opened 2017-02-24
Make cache setup consistent to fix dev profile and dev benchmark-cpu

dylanahsmith:
## Problem

I ran `dev benchmark-cpu` on master and got surprising results, since it looked like hits and misses were taking approximately the same time.  It turns out we setup the cache connection inconsistently, so `dev test` was using identity-cache.railgun as the memcached host and `dev benchmark-cpu` was using localhost. 

## Solution

Move the cache setup into its own file like the database setup, so it will stay consistent across dev/rake tasks.

### DIFF
--- a/performance/cache_runner.rb
+++ b/performance/cache_runner.rb
@@ -6,14 +6,12 @@ require 'identity_cache'
 require 'memcached_store'
 require 'active_support/cache/memcached_store'
 
-$memcached_port = 11211
-$mysql_port = 3306
-
 require File.dirname(__FILE__) + '/../test/helpers/active_record_objects'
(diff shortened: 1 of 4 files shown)


performance/cache_runner.rb
```diff
@@ -6,14 +6,12 @@
 require 'memcached_store'
 require 'active_support/cache/memcached_store'
 
-$memcached_port = 11211
-$mysql_port = 3306
```
daniellaniyo: oh this wasn't used anywhere?

dylanahsmith: Right.  Probably got missed in a previous refactor.

PR #326, opened 2017-03-24
Log all scope values, not just WHERE values

fw42:
Turns out scopes can contain other things besides WHERE.

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -108,7 +108,7 @@ module IdentityCache
 
       def raise_if_scoped
         if current_scope
-          IdentityCache.logger.error("#{name} has scope: #{current_scope.where_values_hash.inspect}")
+          IdentityCache.logger.error("#{name} has scope: #{current_scope.to_sql} (#{current_scope.values.keys})")
           raise UnsupportedScopeError, "IdentityCache doesn't support rails scopes (#{name})"
         end
       end


lib/identity_cache/query_api.rb
```diff
@@ -108,7 +108,7 @@ def prefetch_associations(associations, records)
 
       def raise_if_scoped
         if current_scope
-          IdentityCache.logger.error("#{name} has scope: #{current_scope.where_values_hash.inspect}")
+          IdentityCache.logger.error("#{name} has scope: #{current_scope.values.inspect}")
```
dylanahsmith: What is in `values`?  Could calling inspect on it result in a slow query?  Should we instead just get `current_scope .to_sql` and `current_scope.values.keys` to avoid that possibility?

fw42: I like `to_sql`. Updated.

lib/identity_cache/query_api.rb
```diff
@@ -108,7 +108,7 @@ def prefetch_associations(associations, records)
 
       def raise_if_scoped
         if current_scope
-          IdentityCache.logger.error("#{name} has scope: #{current_scope.where_values_hash.inspect}")
+          IdentityCache.logger.error("#{name} has scope: #{current_scope.to_sql}")
```
dylanahsmith: I would also include `Shop.current_scope.values.keys` since values like `:preload` wouldn't show up.  Note that it is also possible that `unscoped` is being used, which would result in `Shop.current_scope.values.keys` being empty, although an empty current_scope is still significant since it overrides the default scope.

fw42: updated

lib/identity_cache/query_api.rb
```diff
@@ -108,7 +108,7 @@ def prefetch_associations(associations, records)
 
       def raise_if_scoped
         if current_scope
-          IdentityCache.logger.error("#{name} has scope: #{current_scope.where_values_hash.inspect}")
+          IdentityCache.logger.error("#{name} has scope: #{current_scope.to_sql} (#{current_scope.values.keys})")
```
tjoyal: `current_scope.values.keys` does not seems to give out a lot of interesting information?

fw42: Most of the information we will want is in the SQL. Dylan asked for the keys because it will show things like preloading.

dylanahsmith:
> I feel `current_scope.values` would be more useful, but I'm not against it.

@tjoyal do you know if it is safe to inspect those values?  Since inspecting the relation itself will cause it to perform a query and a query like that is likely to be very expensive by just being the base scope.

tjoyal:
I would not know, but I agree it might be risky...

Can you clarify to me what you would be able to understand from only the keys? Some of these keys could point to `nil` or `[]`

eg.:
```
current_scope.values.keys
=> [:order, :from, :lock, :where, :having, :references]
```

dylanahsmith:
> Can you clarify to me what you would be able to understand from only the keys?

What information we are missing, if any, from `to_sql`.  Previously we only included where_values, but that was empty for the case we were debugging.  It could be that current_scope is set by `unscoped`, so if `current_scope.values.keys` returns `[]` then we would know that is the case.

tjoyal:
But we now know that `Model.unscoped.fetch(x)` will raise. 

We can ship the log, but wouldn't the idea to fix idc for it is allowed? (making the log never be hit anymore for that case)

tjoyal:
Ignore my previous comment, `fetch` would consider the default scope, using `unscoped` need to raise.

dylanahsmith:
> But we now know that Model.unscoped.fetch(x) will raise.

The problem we are debugging is that it looks like current_scope is leaking between jobs, likely because of a timeout in one of the jobs.  So we are actually hitting the problem in a case like `Model.fetch_by_id(record_id)`

> We can ship the log, but wouldn't the idea to fix idc for it is allowed?

In this case it is actually good that it raised, since otherwise we could have subtly corrupted the cache.

There is no easy way to support scopes with identity cache, since that would involve executing the SQL in memory on the cached data.  We have no need for the complexity.

PR #330, opened 2017-04-03
Check if cache should be used before expiry

xldenis:
I noticed that it was possible to `expire_cache` on models that have `should_use_cache?` set to `false`. Since this causes a write, it can cause undesirable behaviour. 

cc @Shopify/pods 

TODO:

- [ ] Add tests

I'm not 100% certain this is the correct place to do this in IDC so I'd like  to check before I implement tests.


xldenis:
closing this for now

PR #331, opened 2017-04-05
Allow case insensitive attribute values for secondary indexes.

felix-d:
This PR fixes a simple bug caused my MySQL case insensitivity for nonbinary strings.

#### Disclaimer

It's not really about case sensitivity but about the actual database collation, I'll try to find a better solution than just the `downcase` hack that works in my use case

### The problem

Let's say we have a simple `DiscountCode` model with a secondary index on code `VARCHAR(255)`

```ruby
class DiscountCode < ActiveRecord::Base
   ...
   cache_index :code, unique: true
end
```

This is what happens when fetching with different cases, although bypassing IDC and using `find_by` would work because of MySQL collation.

```ruby
[0] DiscountCode.fetch_by_code("FOO")
=> # cache miss, secondary index created with "FOO"
[1] DiscountCode.create!(code: "foo")
=> # no secondary index "foo" to invalidate
[2] PriceRule::DiscountCode.fetch_by_code("FOO")
=> # cache hit, we can't retrieve the code
```

### The fix

When creating a secondary cache index, we can specify attributes that should be treated as case insensitive

```ruby
class DiscountCode < ActiveRecord::Base
   ...
   cache_index :code, unique: true, case_insensitive: [:code]
end
```

and then...

```ruby
[0] DiscountCode.fetch_by_code("FOO")
=> # cache miss, secondary index created with "foo" because it was downcased
[1] DiscountCode.create!(code: "foo")
=> # secondary index with "foo" invalidated
[2] PriceRule::DiscountCode.fetch_by_code("FOO")
=> # cache miss with "foo", we can retrieve the code! success
```

cc @gmalette

### DIFF
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -41,7 +41,8 @@ module IdentityCache
         raise NotImplementedError, "Cache indexes need an enabled primary index" unless primary_cache_index_enabled
         options = fields.extract_options!
         unique = options[:unique] || false
-        cache_attribute_by_alias('primary_key', 'id', by: fields, unique: unique)
+        case_insensitive_fields = options[:case_insensitive] || []
+        cache_attribute_by_alias('primary_key', 'id', by: fields, unique: unique, case_insensitive: case_insensitive_fields)
 
         field_list = fields.join("_and_")
         arg_list = (0...fields.size).collect { |i| "arg#{i}" }.join(',')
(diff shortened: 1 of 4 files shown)


lib/identity_cache/configuration_dsl.rb
```diff
@@ -262,6 +265,15 @@ def attribute_dynamic_fetcher(attribute, fields, values, unique_index) #:nodoc:
         end
       end
 
+      def transform_case_insensitive_values(fields, values, case_insensitive_fields)
+        return values if case_insensitive_fields.empty?
+
+        fields.zip(values).map do |field, value|
+          next(value.downcase) if case_insensitive_fields.include?(field)
```
simon-shopify: What's the collation on the fields you are trying to index? I think the comment you added to the README is very misleading.

felix-d: What's the collation on the fields you are trying to index?

`utf8mb4_general_ci`

Downcasing is probably too simple and a configurable encoding should be used instead? WDYT?

felix-d:
Closing, until I have a solution, maybe a custom encoding lambda instead

PR #332, opened 2017-04-06
Allow attribute value preprocessing at secondary index cache key generation

felix-d:
This PR offers a workaround for a simple bug caused by database collation.

### Disclaimer

I am not convinced this is the best approach... at all. 

Another solution would be encoding at the database level, but this has some downsides.

### The problem

Let's say we have a simple `DiscountCode` model with a secondary index on code, which is a `VARCHAR(255) ` with `utf8mb4_general_ci` collation.

```ruby
class DiscountCode < ActiveRecord::Base
   ...
   cache_index :code, unique: true
end
```

This is what happens when fetching with different cases, although using `find_by` would work because of MySQL collation.

```ruby
[0] DiscountCode.fetch_by_code("FOO")
=> # cache miss, secondary index created with "FOO"
[1] DiscountCode.create!(code: "foo")
=> # no secondary index "foo" to invalidate
[2] PriceRule::DiscountCode.fetch_by_code("FOO")
=> # cache hit, we can't retrieve the code
```

### The fix

When creating a secondary cache index, we can specify attributes that should be preprocessed.

```ruby
class DiscountCode < ActiveRecord::Base
   ...
   cache_index :code, unique: true, prepare: { code: ->(code) { code.downcase }}
end
```

and then...

```ruby
[0] DiscountCode.fetch_by_code("FOO")
=> # cache miss, secondary index created with "foo" because it was downcased
[1] DiscountCode.create!(code: "foo")
=> # secondary index with "foo" invalidated
[2] PriceRule::DiscountCode.fetch_by_code("FOO")
=> # cache miss with "foo", we can retrieve the code! success
```


felix-d:
This is a cool workaround, but maybe this should be fixed with an encoding at the database level, not in IDC.

For example, before creating a DiscountCode you could encode the code

```ruby
Base32::Crockford.normalize(code) 
```
and store that instead, alongside the raw value in another column for display and reporting purposes, and use the encoded column for fetching.

PR #335, opened 2017-05-01
Remove deprecated code

dylanahsmith:
Since the next release will have breaking changes, I thought we should any deprecated code.

This PR removes the disable_primary_cache_index method and the `embed: false` cache_has_many option.


dylanahsmith:
Sure, we could make the next version a 1.0 release.

PR #337, opened 2017-05-01
Add rails 5.1 to CI

dylanahsmith:


.travis.yml
```diff
@@ -7,11 +7,14 @@ rvm:
...
   - Gemfile.rails42
   - Gemfile.rails50
+  - Gemfile.rails51
 
 matrix:
   exclude:
     - gemfile: Gemfile.rails50
       env: DB=postgresql
+    - gemfile: Gemfile.rails51
+      env: DB=postgresql
```
dylanahsmith: This has the same test failure as rails 5.0 for an unintended serialization format change

test/schema_change_test.rb
```diff
@@ -1,13 +1,15 @@
 require "test_helper"
 
 class SchemaChangeTest < IdentityCache::TestCase
-  class AddColumnToChild < ActiveRecord::Migration
+  Migration = ActiveRecord::Migration.try(:[], 4.2)
```
rafaelfranca: Would not this return `nil`?

dylanahsmith: Oops, meant to default to `ActiveRecord::Migration`.  Pushed a fix

Issue #338, opened 2017-05-16
Association not working correctly with custom primary keys


jstastny:
Assuming the following models (custom primary keys are used instead of the default `id`):

```ruby
class CustomMasterRecord < ActiveRecord::Base
  include IdentityCache

  has_many :custom_child_record, foreign_key: :master_id
  cache_has_many :custom_child_record
  self.primary_key = 'master_primary_key'
end

class CustomChildRecord < ActiveRecord::Base
  include IdentityCache
  belongs_to :custom_master_record, foreign_key: :master_id
  cache_belongs_to :custom_master_record
  self.primary_key = 'child_primary_key'
end
```

The following will fail:

```ruby
  CustomMasterRecord.fetch 1
```

```
ActiveRecord::StatementInvalid: PG::UndefinedColumn: ERROR:  column "master_primary_key" does not exist
LINE 1: SELECT "custom_child_records"."master_id", master_primary_ke...
                                                   ^
: SELECT "custom_child_records"."master_id", master_primary_key FROM "custom_child_records" WHERE "custom_child_records"."master_id" = 1
```

I already have fix prepared for this. PR coming shortly.

PR #339, opened 2017-05-16
Fix the column name in association query

jstastny:
This didn't work correctly for the models using custom primary key.

Fixes https://github.com/Shopify/identity_cache/issues/338

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -228,7 +228,7 @@ module IdentityCache
           scope = child_model.all
           scope = scope.instance_exec(nil, &reflection.scope) if reflection.scope
 
-          pairs = scope.where(reflection.foreign_key => records.map(&:id)).pluck(reflection.foreign_key, reflection.active_record_primary_key)
+          pairs = scope.where(reflection.foreign_key => records.map(&:id)).pluck(reflection.foreign_key, reflection.association_primary_key)
           ids_by_parent = Hash.new{ |hash, key| hash[key] = [] }
           pairs.each do |parent_id, child_id|
             ids_by_parent[parent_id] << child_id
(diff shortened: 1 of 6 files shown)


lib/identity_cache/query_api.rb
```diff
@@ -228,7 +228,7 @@ def preload_id_embedded_associations(records)
           scope = child_model.all
           scope = scope.instance_exec(nil, &reflection.scope) if reflection.scope
 
-          pairs = scope.where(reflection.foreign_key => records.map(&:id)).pluck(reflection.foreign_key, reflection.active_record_primary_key)
+          pairs = scope.where(reflection.foreign_key => records.map(&:id)).pluck(reflection.foreign_key, child_model.primary_key)
```
camilo: @rafaelfranca this looks okay to me, but can you think of how this can backfire?

rafaelfranca: `reflection.active_record_primary_key` call `child_model.primary_key` but it also respect the `primary_key` option in the association definition and I think it should. @jstastny why should we ignore that value?

jstastny: I am not sure I fully understand what you are saying, but for the newly create test case, we get following:
- `scope.klass` = `CustomChildRecord`
- `reflection.foreign_key` = `master_id`
- `reflection.active_record_primary_key` = `master_primary_key`

If we run the original code, it means something like `SELECT master_id, master_primary_key FROM custom_child_records`. This ends up with error, as `master_primary_key` column does not exist in the child table. Therefore I replaced it with `child_model.primary_key`, which is what I think should be here.

The reason it worked so far is in my opinion that it relied on the fact both the child and master tables used `id` as the primary key and therefore both `reflection.active_record_primary_key` and `child_model.primary_key` returned `id`.

rafaelfranca: Ok. Just saw the problem. We are using `reflection.active_record_primary_key` that is getting the primary key of the associated object. What we want here is the `reflection.association_primary_key` that will get the primary key of the reflection class but still respect the `:primary_key` option. Can you change it?

jstastny: Oh I see. Thanks for the comment. Changed in 69e47025ea891ba327f0e4015c5ee42075811f3d. Do you want me to rebase to get rid of the extra commit?

rafaelfranca:
Thanks!

PR #340, opened 2017-05-19
Adding ActiveSupport::Notifications for (de)hydration

alexsnaps:
Adding `ActiveSupport::Notifications` around hydration & dehydration of records.
Enabling to gain better insights on how IDC performs, measuring latency of these operations

### DIFF
--- a/lib/identity_cache/memoized_cache_proxy.rb
+++ b/lib/identity_cache/memoized_cache_proxy.rb
@@ -30,34 +30,47 @@ module IdentityCache
     end
 
     def write(key, value)
-      memoized_key_values[key] = value if memoizing?
-      @cache_fetcher.write(key, value)
+      memoizing = memoizing?
+      ActiveSupport::Notifications.instrument('identity_cache.backend.write', memoizing: memoizing) do
+        memoized_key_values[key] = value if memoizing
+        @cache_fetcher.write(key, value)
+      end
     end
 
     def delete(key)
-      memoized_key_values.delete(key) if memoizing?
-      result = @cache_fetcher.delete(key)
-      IdentityCache.logger.debug { "[IdentityCache] delete #{ result ? 'recorded'  : 'failed'  } for #{key}" }
-      result
+      memoizing = memoizing?
+      ActiveSupport::Notifications.instrument('identity_cache.backend.delete', memoizing: memoizing) do
+        memoized_key_values.delete(key) if memoizing
+        result = @cache_fetcher.delete(key)
+        IdentityCache.logger.debug {"[IdentityCache] delete #{ result ? 'recorded' : 'failed'  } for #{key}"}
+        result
+      end
     end
 
     def fetch(key)
       used_cache_backend = true
       missed = false
-      value = if memoizing?
-        used_cache_backend = false
-        memoized_key_values.fetch(key) do
-          used_cache_backend = true
-          memoized_key_values[key] = @cache_fetcher.fetch(key) do
+      memoizing = memoizing?
+      value = ActiveSupport::Notifications.instrument('identity_cache.backend.fetch', { memoizing: memoizing }) do |payload|
+        value = if memoizing
+          used_cache_backend = false
+          memoized_key_values.fetch(key) do
+            used_cache_backend = true
+            memoized_key_values[key] = @cache_fetcher.fetch(key) do
+              missed = true
+              yield
+            end
+          end
+        else
+          @cache_fetcher.fetch(key) do
             missed = true
             yield
           end
         end
-      else
-        @cache_fetcher.fetch(key) do
-          missed = true
-          yield
-        end
+        payload[:memoized] = used_cache_backend ? 0 : 1
+        payload[:miss] = missed ? 1 : 0
+        payload[:populated] = (missed && value) ? 1 : 0
+        value
       end
 
       if missed
@@ -72,33 +85,47 @@ module IdentityCache
     def fetch_multi(*keys)
       memoized_keys, missed_keys = [], [] if IdentityCache.logger.debug?
 
-      result = if memoizing?
-        hash = {}
-        mkv = memoized_key_values
+      miss_count = 0
+      memoized_count = 0
 
-        non_memoized_keys = keys.reject do |key|
-          if mkv.has_key?(key)
-            memoized_keys << key if IdentityCache.logger.debug?
-            hit = mkv[key]
-            hash[key] = hit unless hit.nil?
-            true
+      memoizing = memoizing?
+      result = ActiveSupport::Notifications.instrument('identity_cache.backend.fetch_multi', { memoizing: memoizing }) do |payload|
+        result = if memoizing
+          hash = {}
+          mkv = memoized_key_values
+
+          non_memoized_keys = keys.reject do |key|
+            if mkv.has_key?(key)
+              memoized_keys << key if IdentityCache.logger.debug?
+              hit = mkv[key]
+              hash[key] = hit unless hit.nil?
+              true
+            end
           end
-        end
 
-        unless non_memoized_keys.empty?
-          results = @cache_fetcher.fetch_multi(non_memoized_keys) do |missing_keys|
+          memoized_count = keys.length - non_memoized_keys.length
+
+          unless non_memoized_keys.empty?
+            results = @cache_fetcher.fetch_multi(non_memoized_keys) do |missing_keys|
+              miss_count = missing_keys.length
+              missed_keys.concat(missing_keys) if IdentityCache.logger.debug?
+              yield missing_keys
+            end
+            mkv.merge! results
+            hash.merge! results
+          end
+          hash
+        else
+          @cache_fetcher.fetch_multi(keys) do |missing_keys|
+            miss_count = missing_keys.length
             missed_keys.concat(missing_keys) if IdentityCache.logger.debug?
             yield missing_keys
           end
-          mkv.merge! results
-          hash.merge! results
         end
-        hash
-      else
-        @cache_fetcher.fetch_multi(keys) do |missing_keys|
-          missed_keys.concat(missing_keys) if IdentityCache.logger.debug?
-          yield missing_keys
-        end
+        payload[:memoized] = memoized_count
+        payload[:miss] = miss_count
+        payload[:populated] = miss_count - (keys.length - result.length)
+        result
       end
 
       log_multi_result(memoized_keys, keys - missed_keys - memoized_keys, missed_keys) if IdentityCache.logger.debug?
@@ -107,8 +134,10 @@ module IdentityCache
     end
 
     def clear
-      clear_memoization
-      @cache_fetcher.clear
+      ActiveSupport::Notifications.instrument('identity_cache.backend.clear') do
+        clear_memoization
+        @cache_fetcher.clear
+      end
     end
 
     private
@@ -118,7 +147,7 @@ module IdentityCache
     end
 
     def memoizing?
-      Thread.current[:memoizing_idc]
+      !!Thread.current[:memoizing_idc]
     end
 
     def log_multi_result(memoized_keys, backend_keys, missed_keys)
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -121,6 +121,13 @@ module IdentityCache
       end
 
       def record_from_coder(coder) #:nodoc:
+        return unless coder
+        ActiveSupport::Notifications.instrument('identity_cache.hydration', class: coder[:class]) do
+          record_from_coder_internal(coder)
+        end
+      end
+
+      def record_from_coder_internal(coder) #:nodoc:
         if coder
           klass = coder[:class].constantize
           record = klass.instantiate(coder[:attributes].dup)
@@ -148,11 +155,11 @@ module IdentityCache
         value = if IdentityCache.unmap_cached_nil_for(coder_or_array).nil?
           nil
         elsif (reflection = record.class.reflect_on_association(association_name)).collection?
-          associated_records = coder_or_array.map {|e| record_from_coder(e) }
+          associated_records = coder_or_array.map {|e| record_from_coder_internal(e) }
           set_inverse_of_cached_has_many(record, reflection, associated_records)
           associated_records
         else
-          record_from_coder(coder_or_array)
+          record_from_coder_internal(coder_or_array)
         end
         variable_name = record.class.send(:recursively_embedded_associations)[association_name][:records_variable_name]
         record.instance_variable_set(:"@#{variable_name}", value)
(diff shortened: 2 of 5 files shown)


lib/identity_cache/query_api.rb
```diff
@@ -120,7 +120,14 @@ def check_association_scope(association_name)
         end
       end
 
-      def record_from_coder(coder) #:nodoc:
+      def record_from_coder(coder)
```
dylanahsmith: I thought the point of the instrumentation was to we don't need to make something like this documented/stable.  It seems like this should still be undocumented.  Maybe instead this should be called `instrumented_record_from_coder` so we don't need to name the method it calls `record_from_coder_internal` which would make this method sound like it isn't internal.

lib/identity_cache/query_api.rb
```diff
@@ -120,7 +120,14 @@ def check_association_scope(association_name)
...
+        return unless coder
+        ActiveSupport::Notifications.instrument('identity_cache.hydration', class: coder[:class]) do
```
dylanahsmith: I hadn't heard the term hydration before.  I found a [What does it mean to hydrate an object?](http://stackoverflow.com/questions/6991135/what-does-it-mean-to-hydrate-an-object) stack overflow answer where I found this part of the explanation helpful

> deserialization == instantiation + hydration

In our case we are also instantiating the active record object as well as "hydrating" it, so it seems like deserialization is the more appropriate term.  That's good because deserialization is also the more familiar term to me.

lib/identity_cache/query_api.rb
```diff
@@ -161,13 +168,20 @@ def set_embedded_association(record, association_name, coder_or_array) #:nodoc:
...
+          embedded_variable.map {|e| coder_from_record_internal(e) }
         else
-          coder_from_record(embedded_variable)
+          coder_from_record_internal(embedded_variable)
         end
       end
 
       def coder_from_record(record) #:nodoc:
+        return unless record
+        ActiveSupport::Notifications.instrument('identity_cache.dehydration', class: record.class.name) do
```
dylanahsmith: Similar to how I think we should use `deserialization` instead of `hydration`, I think this should be `serialization`

lib/identity_cache/query_api.rb
alexsnaps: Actually `hydration` comes from the ORM world, where what's meant is "hydrating" the actual Object from a the "split up representation" in this case, the `Hash`. Where as serialization would be getting that "dehydrated" representation into a `byte[]`

That's why I don't like using serialization as the wording, but if it's "better accepted/known" I don't really care so much either. Whatever the consensus is :)

lib/identity_cache/query_api.rb
alexsnaps: See above.

lib/identity_cache/query_api.rb
alexsnaps: i.e. serialization actually happens within the `MemcachedStore`layer (iirc) where the `Hash` gets serialized into a byte array. This process _is not part_ of this measurement

alexsnaps: having read the SO link now... I pretty much agree with that [comment though](http://stackoverflow.com/questions/6991135/what-does-it-mean-to-hydrate-an-object/20787106#comment31159656_6991192):
>Hydration refers to filling an existing object with data. It is not a synonym for serialization

Interesting discussions further down too. Also "ActiveRecord hydrating" doesn't seem to foreign to the "Rails community" either. In any case... what ever makes it the clearest to people consuming this... but again, this is not accounting for "serialization time before going over the wires" (which could be 0 with say an in-memory backed to IDC, yet (de)hydration would still be required). I hope I made it somewhat clearer what _this_ metric is about. Let me know what you guys think is the best naming though!

dylanahsmith: good point, deserialization would make it sounds like it includes more and thus would be more inaccurate than hydration

alexsnaps:
>dylanahsmith [9 minutes ago](https://github.com/Shopify/identity_cache/pull/340#discussion_r117510391) Member
good point, deserialization would make it sounds like it includes more and thus would be more inaccurate than hydration

So we're good? Or can you think of some other term that would avoid that conflation? 
As a side note, I was thinking of adding further AS notifications for when we're actually hitting the underlying cache (hopefully being able to report misses, as these then also populate and include underlying SoR latency), so anyone could get a complete performance profile. So in the case of a hit, that metric would then include serialization. 

Anyways, the point being, I think it'd be nice to provide a complete picture of how the IDC behaves, and with that in mind (if you guys agree to the idea obviously), I was hoping naming _may_ be somewhat easier :)

alexsnaps:
I could also postpone merging this, while working on the other AS notifications btw... Maybe having these actual changes in code, make it easier for all of us to reason about names? Again, if we all agree the additional AS notifications make sense...

rafaelfranca:
Yeah, maybe we should merge all them together. I think those additional AS notifications make sense. If we have a need to measure something we should expose them using AS notifications otherwise we would have to monkey patch. Since the instrumentations are cheap (if you don't subscribe they have no overhead at all) it is a good idea to add them and avoid monkey patches.

alexsnaps:
👍  Cool, I'll work on them!

alexsnaps:
I've added some additional stats... Now I wonder if I should have "nested" instrumentations for "hitting the database" metrics... though, otoh, these are maybe best added at the db layer's level anyways. Anyways, let me know what you guys think. I'll be adding tests accordingly.

lib/identity_cache/memoized_cache_proxy.rb
```diff
@@ -30,34 +30,47 @@ def with_memoization(&block)
...
     def fetch(key)
       used_cache_backend = true
       missed = false
-      value = if memoizing?
-        used_cache_backend = false
-        memoized_key_values.fetch(key) do
-          used_cache_backend = true
-          memoized_key_values[key] = @cache_fetcher.fetch(key) do
+      memoizing = memoizing?
+      value = ActiveSupport::Notifications.instrument('identity_cache.backend.fetch', { memoizing: memoizing }) do |payload|
```
dylanahsmith: Does it make sense to instrument how long it takes to fetch a memoized value?

lib/identity_cache/memoized_cache_proxy.rb
```diff
@@ -30,34 +30,47 @@ def with_memoization(&block)
...
           end
         end
-      else
-        @cache_fetcher.fetch(key) do
-          missed = true
-          yield
-        end
+        payload[:memoized] = used_cache_backend ? 0 : 1
+        payload[:miss] = missed ? 1 : 0
+        payload[:populated] = (missed && value) ? 1 : 0
```
dylanahsmith: This looks like a C-like pattern for booleans.  Why not just use ruby boolean values?

dylanahsmith: Oh, nevermind, I can see numeric values makes sense for the batched case.

lib/identity_cache/query_api.rb
```diff
@@ -121,6 +121,13 @@ def check_association_scope(association_name)
       end
 
       def record_from_coder(coder) #:nodoc:
+        return unless coder
+        ActiveSupport::Notifications.instrument('identity_cache.hydration', class: coder[:class]) do
+          record_from_coder_internal(coder)
+        end
+      end
+
+      def record_from_coder_internal(coder) #:nodoc:
```
dylanahsmith: As already mentioned in https://github.com/Shopify/identity_cache/pull/340#discussion_r117493405, I don't like calling this internal, since it makes it sound like `record_from_coder` isn't internal

lib/identity_cache/memoized_cache_proxy.rb
alexsnaps: Right, I wanted consistency between the two.

lib/identity_cache/query_api.rb
alexsnaps: Y, somehow missed that rename, I'll do that!

lib/identity_cache/memoized_cache_proxy.rb
alexsnaps: Felt like this should be fast enough given there is no contention on these... But that could be added as well. So are you arguing for "nested" statistics? (i.e. the whole thing took n, but is split in between: memoized hits, cache hits, db hits?
Should the last one be implemented at this level though? Or do we leave that to the "block" to do?

dylanahsmith: What I'm saying is that you **are** instrumenting the case where it is fetched an already memoized value.  The problem is that it would be fast and thus could easily skew the statistics to make it seem like a cache fetch against memcached is much faster than it actually is.

alexsnaps: Right, I get that I am... Currently these `#fetch(_multi)` measure (up to) 3 things:
 1. memoized look up (if memoizing, whether hit or not, that's a HashTable lookup); then, in case of miss
 1. memcached look up (again, that's a network hit, plus the actual look up); then, in case of miss again
 1. we hit the SoR... which again could "miss".

Now the issue being that, given the "cache-through pattern", we may or may not hit either tier (threadlocal, or memcached) and even maybe even the database in a single of these metrics; I see we have a couple of choices: 

 1. Not measure at the "outer most" layer and measure each of these 3 separately (though I'm not sure the SoR time should be measured here in that case; so I'm arguing for thread-local and/or memcached); or
 1. Measure the outer layer, but measure time spent in each of the tiers and publish all of that information in the `payload` (i.e. what's there, plus the actual times); or
 1. Measure the outer layer, plus each individual layers ("nested stats"); or
 1. What I have there, only the overall time spent, plus a break down.
 1. Any other option?

I wasn't happy about option 3 (the current one), so that's the discussion I wanted to trigger... but wasn't probably very clear. wdyt?

alexsnaps: Having slept over this... I think, I'm voting for measuring one huge aggregate and publishing individual "times" (along side the current "per tier" metrics) as part of the payload. 
The reasoning being that I'd have to measure the `yield` block's execution time anyways, in the case of a miss (and have the caller subtract it himself, if there would be a `@cache_fetcher.fetch` notification); so I think we're better off _not_ throwing that piece of information away and just publish a notification, with the additional information of how time was spent:
 * hitting the memoizing layer
 * hitting memcached 
 * hitting the actual SoR

Given the API of the AS::Notification and the patterns used by IDC, I think that makes the most sense. wdyt?

alexsnaps:
Here is a more concrete example of what I'm proposing:
```ruby

# measure the thing head to tail
result = ActiveSupport::Notifications.instrument('identity_cache.backend.fetch_multi', { memoizing: memoizing }) do |payload|
  result = if memoizing
    hash = {}
    mkv = memoized_key_values
 
    # expose the time spent in the threadlocal storage 
    payload[:time_memoizing] = Benchmark.realtime do
      non_memoized_keys = keys.reject do |key|
        if mkv.has_key?(key)
          memoized_keys << key if IdentityCache.logger.debug?
          hit = mkv[key]
          hash[key] = hit unless hit.nil?
          true
        end
      end
    end
    memoized_count = keys.length - non_memoized_keys.length
    unless non_memoized_keys.empty?
      time_in_cache = Benchmark.realtime do
        results = @cache_fetcher.fetch_multi(non_memoized_keys) do |missing_keys|
          miss_count = missing_keys.length
          missed_keys.concat(missing_keys) if IdentityCache.logger.debug?

          # expose the time spent in the underlying System of Records (SoR) 
          payload[:time_sor] = Benchmark.realtime do
            yield missing_keys
          end
        end
        mkv.merge! results
        hash.merge! results
      end

      # expose the time spent in memcached (both when hitting, as well as populating)
      payload[:time_in_cache] = time_in_cache - payload[:time_sor]
    end
    hash
  else
    # SNIP
  end
  payload[:hit_memoizing] = memoized_count

  # gives more context on the `:time_in_cache` metric
  payload[:hit_cache] = keys.length - miss_count - memoized_count
  payload[:populated] = miss_count - (keys.length - result.length)

  # really a miss on the SoR, will be memoized and/or cached from there on.
  payload[:absent_from_sor] = keys.length - result.length
  result
end
```

dylanahsmith:
I still don't see why you care about the time spent memoizing.  It is just a hash table lookup, which isn't going to be significant.  This is unlike a cache hit which actually has to go to memcached.

Why don't you just move the instrumentation to CacheFetcher to avoid instrumenting the time spent memoizing?

alexsnaps:
We wouldn't have any metrics for things "only" hitting the memoizing layer then... Unsure how I feel about that. Feels like underplaying the cache's performance. This just reflects all that's going on, and one can just ignore notifications where `payload[:hit_cache] == 0 && payload[:populated]` if one so desires... That also helps one seeing whether memoizing is even useful (say if `payload[:hit_memoizing] == 0` "most" of the time). My idea is to provide a user with all the insight he may desire on how IDC is performing for him.
Now we could not measure `payload[:time_memoizing]` and let one infer it's `overall_time - time_in_cache - time_sor` ...

dylanahsmith:
There is still plenty of memoization happening at the model which will get used without going to MemoizedCacheProxy, so we wouldn't really be seeing the whole picture.  For instance, we would like to decouple the Shopify codebase more, which will involve removing some associations and relying on this memoization more (e.g. `record.fetch_foo` -> `Foo.fetch(record.foo_id)`).  That would cause this stat to increase, when that noise doesn't matter.

alexsnaps:
>that noise doesn't matter.
What noise? The overhead imposed by the memoization (i.e. that it's part of `identity_cache.backend.fetch_multi` measured duration)? Not sure I understand _what_ you are trying to solve. I can see getting rid of the  `:time_memoizing` metric as a good thing, so I'll remove that (as it can be inferred, no need to add it explicitly and incur the overhead of measuring it explicitly). Is that enough to address your concern; if not, what's the concern?

dylanahsmith:
By noise I mean instrumentation noise.  It doesn't provide us value to remove those, and the debug logging for it is probably something else we will want to get rid of to avoid noisey logs in development.

alexsnaps:
Cool, then we understand each other, but it's _not_ noise, it's actually everything I'm currently trying to measure. I want to know when we "only" hit the memoization and how much the layer actually "saves". I'll just remove the explicit timing of the `payload[:time_memoizing]` as in the above example, as discusses, as this isn't necessary. The goal is to figure out how we best proceed with our caching strateg(y|ies).
I'll wrap this PR up and add tests and I'll then force push a meaningful set of commits on this branch.

csfrancis:
I tend to agree with @dylanahsmith here. 

> Moving the entire instrumentation, we wouldn't have any metrics for things "only" hitting the memoizing layer then... Unsure how I feel about that. Feels like underplaying the cache's performance.

Like Dylan was saying, a memoized hit is basically just a hash table lookup - which to me seems free, relative to having to go to memcached or MySQL. It also seems like having the developer filter for these events (`payload[:hit_cache] == 0 && payload[:populated]`) will make the developer have to do more work to get non-skewed numbers (and developers are lazy).

csfrancis:
Chatted with @alexsnaps on Slack, and I do think that this approach works well if we truly want to observe as much as we can. I think a legitimate argument for leaving the memoization metrics there is to see if it's even worthwhile to use in the first place.

👍

alexsnaps:
>developers are lazy

... and that's a good thing, that's why I added the `: time_in_cache` & `:time_in_sor` directly as part of the `payload` there for them to use (plus the actual "hit counts", so you _could_ even do averages on the `_multi` if one would want to).

Anyways, I'll just close this and open a new one with all the actual changes and we can take it from there...

PR #341, opened 2017-05-25
Hydration & cache ops notifications

alexsnaps:
Follow up to https://github.com/Shopify/identity_cache/pull/340

I left the naming as `identity_cache.(de)hydration`, added `identity_cache.cache.(write|delete|clear|fetch|fetch_all)` notifications. 
The `fetch(_all)` are ones we had internally monkey patching the `MemoizedCacheProxy` class, subscribing to these would be the exact same thing now. The payload of such a notification contains additional information, such as: hit count to the different layers, plus how much was populated in the cache. `:time_in_sor` is the time spent from the Cache's perspective in the `yield` block (so this would also include dehydration time for instance). 

For "finer" instrumentation of how the underlying cache behaves, both `MemcachedStore` and `ActiveSupport::Cache::Store` provide additional instrumentation metrics (e.g. read vs. write latency).

I hope this makes sense. I hope this addresses everyone's concerns.

### DIFF
--- a/lib/identity_cache/memoized_cache_proxy.rb
+++ b/lib/identity_cache/memoized_cache_proxy.rb
@@ -1,4 +1,5 @@
 require 'monitor'
+require 'benchmark'
 
 module IdentityCache
   class MemoizedCacheProxy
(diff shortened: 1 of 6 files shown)


lib/identity_cache/memoized_cache_proxy.rb
```diff
@@ -37,34 +38,58 @@ def with_memoization(&block)
     end
 
     def write(key, value)
-      memoized_key_values[key] = value if memoizing?
-      @cache_fetcher.write(key, value)
+      memoizing = memoizing?
+      ActiveSupport::Notifications.instrument('identity_cache.cache.write', memoizing: memoizing) do
```
csfrancis: I'd prefer it if the AS notifications were decorator methods, but I can't think of a clean way to do it that would work with `fetch` and `fetch_multi`.

lib/identity_cache/memoized_cache_proxy.rb
```diff
@@ -79,33 +104,56 @@ def fetch(key)
...
-            hit = mkv[key]
-            hash[key] = hit unless hit.nil?
-            true
+      miss_count = 0
+      memoized_count = 0
+
+      memoizing = memoizing?
+      result = ActiveSupport::Notifications.instrument('identity_cache.cache.fetch_multi', { memoizing: memoizing }) do |payload|
+        payload[:time_in_sor] = 0
+        result = if memoizing
```
dylanahsmith: Similar to my suggestion to extract out this branch on `fetch`, I think we can do the same here, where we can extract out even more logic

```ruby
    def fetch_multi_memoized(keys)
      return yield keys unless memoizing?

      result = {}
      missing_keys = keys.reject do |key|
        if memoized_key_values.key?(key)
          result[key] = memoized_key_values[key]
          true
        end
      end

      unless missing_keys.empty?
        block_result = yield missing_keys
        memoized_key_values.merge!(block_result)
        result.merge!(block_result)
      end

      result
    end
```

alexsnaps:
@dylanahsmith I really like your refactoring pass on this! I take all the blame, I should have had started with a refactoring first before adding all my mess... I really appreciate you doing this!

lib/identity_cache/memoized_cache_proxy.rb
```diff
@@ -37,101 +38,152 @@ def with_memoization(&block)
...
+      cache_misses = 0
+
+      value = ActiveSupport::Notifications.instrument('identity_cache.cache.fetch') do |payload|
+        payload[:resolve_miss_time] = 0.0
+
+        value = fetch_memoized(key) do
+          memo_misses = 1
+          @cache_fetcher.fetch(key) do
+            cache_misses = 1
+            instrument_duration(payload, :resolve_miss_time) do
```
rafaelfranca: Why are we doing the duration calculation here? The event generated by ActiveSupport::Notifications already have the duration. If we need two different timing for the same instrumentation maybe we need two different instrumentations?

alexsnaps: The notification will provide the subscriber with start & finish time of the _whole_ thing. This duration measured here is a couple of things: hitting the underlying SoR, dehydrating the AR result and populating IDC for all keys.

I find that having this provides a complete picture of what IDC is doing. If it'd be a different notification, you'd need to compose the result from the both notifications to have that same picture. Now I am tainted as I'm interested in knowing how much IDC "costs" us. 

I'm not even sure what the pattern would be to compose notifications, a thread local? :(
As one may still want to subscribe to `ActiveSupport::Cache::Store` notifications to split read vs. write cost of cache ops

rafaelfranca: Notifications can be nested. So:

```
instrument("idc.fetch") do
  #do some things

  instrument("idc.resolve_miss_time") do
    # do some things
  end
end
```

If you want to know the duration of the entire fetch it is possible by reading the `idc.fetch`. If you want to know the duration of the miss it is possible by reading the `"idc.resolve_miss_time"`.

Does not that work for you use case?

alexsnaps: I get the notification end... the subscribe is more what confuses me.
If I want to (programmatically) subtract the inner instrumentation's duration (`idc.resolve_miss_time` in your example) from the outer, I'd need to store it in a threadlocal and retrieve it back in the `idc.fetch`'s subscriber. Feels kinda yuk to me... 

Anyways, I guess you're in favor of two instrumentations then, right? If this feels better to everyone, I'm good with it.

PR #345, opened 2017-06-16
Set inverse cached association for cache_has_one on cache hit

dylanahsmith:
## Problem

For a `cache_has_many` we were setting the inverse cached association for associated records, but we weren't doing the same thing for `cache_has_one`.

## Solution

I generalized `set_inverse_of_cached_has_many` into `set_inverse_of_cached_association` which no longer assumes that the cached association options are in `cached_has_manys` or that the association target is an array.

To make it easier to get the cached association options across associations types, I've added `cached_association_options` which is used in `set_embedded_association`.  I've also added a `:prepopulate_method_name` cached association option key and associated method for recursively embedded associations to avoid yet another inconsistency.

I have also separated out the step to turn the coder objects into their corresponding active record objects to keep `set_embedded_association` more focused.

The added regression test fails without the other changes in this PR.

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -132,35 +132,48 @@ module IdentityCache
           klass = coder[:class].constantize
           record = klass.instantiate(coder[:attributes].dup)
 
-          coder[:associations].each {|name, value| set_embedded_association(record, name, value) } if coder.has_key?(:associations)
+          coder[:associations].each do |name, value|
+            set_embedded_association(record, name, hydrate_association_target(value))
+          end if coder.has_key?(:associations)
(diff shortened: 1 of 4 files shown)


lib/identity_cache/query_api.rb
```diff
@@ -132,35 +132,48 @@ def record_from_coder(coder) #:nodoc:
...
+      def hydrate_association_target(dehydrated_value)
+        dehydrated_value = IdentityCache.unmap_cached_nil_for(dehydrated_value)
+        if dehydrated_value.is_a?(Array)
+          dehydrated_value.map { |coder| record_from_coder(coder) }
+        else
+          record_from_coder(dehydrated_value)
+        end
+      end
+
+      def set_inverse_of_cached_association(record, association_options, association_target)
```
rafaelfranca: Just to make the class easier to review I'd move this method definition to after `set_embedded_association` since it is were it is used.

PR #348, opened 2017-07-19
Add support for blob replication.

simon-shopify:
An access pattern we see is when a sigle row becomes really hot. Because of the way memcached works, all the load goes to a single memcached server, which we have seen become slow.

This PR does not fix the issue, but mitigates it by replicating blobs under multiple keys. For 2 replicas, there is a (_n_ - 1) / _n_ chance they fall on different servers and the load is split, given a memcached ring of _n_ servers.

An alternative would be to do the replication right in the memcached client, as the distribution is computed so we could pick instead the _n_ + _k_ (mod _n_) th server instead of a random one to make sure they land on different machines.

@csfrancis, @sirupsen is that what you had in mind?

### DIFF
--- a/lib/identity_cache.rb
+++ b/lib/identity_cache.rb
@@ -15,6 +15,7 @@ require "identity_cache/cache_invalidation"
 require "identity_cache/cache_fetcher"
 require "identity_cache/fallback_fetcher"
 require 'identity_cache/without_primary_index'
+require "identity_cache/replicated_cache_proxy"
 
 module IdentityCache
   extend ActiveSupport::Concern
@@ -55,6 +56,9 @@ module IdentityCache
     mattr_accessor :fetch_read_only_records
     self.fetch_read_only_records = true
 
+    mattr_accessor :blob_replication_factor
+    self.blob_replication_factor = 1
+
     def included(base) #:nodoc:
       raise AlreadyIncludedError if base.respond_to?(:cached_model)
       base.class_attribute :cached_model
@@ -72,12 +76,12 @@ module IdentityCache
       if defined?(@cache)
         cache.cache_backend = cache_adaptor
       else
-        @cache = MemoizedCacheProxy.new(cache_adaptor)
+        @cache = wrap_in_replicated_proxy(MemoizedCacheProxy.new(cache_adaptor))
       end
     end
 
     def cache
-      @cache ||= MemoizedCacheProxy.new
+      @cache ||= wrap_in_replicated_proxy(MemoizedCacheProxy.new)
     end
 
     def logger
@@ -160,6 +164,14 @@ module IdentityCache
       keys.each_slice(BATCH_SIZE).each_with_object Hash.new do |slice, result|
         result.merge! cache.fetch_multi(*slice) {|missed_keys| yield missed_keys }
       end
+    end
+
+    def wrap_in_replicated_proxy(backend)
+      return backend if blob_replication_factor == 1
+      ReplicatedCacheProxy.new(
+        backend,
+        blob_replication_factor: blob_replication_factor,
+      )
     end
   end
 end
--- a/lib/identity_cache/replicated_cache_proxy.rb
+++ b/lib/identity_cache/replicated_cache_proxy.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+module IdentityCache
+  class ReplicatedCacheProxy
+    def initialize(cache_backend, blob_replication_factor:)
+      @cache_backend = cache_backend
+      @blob_replication_factor = blob_replication_factor
+    end
+
+    def write(key, value)
+      keys(key).each do |key|
+        cache_backend.write(key, value)
+      end
+    end
+
+    def delete(key)
+      keys(key).each do |key|
+        cache_backend.delete(key)
+      end
+    end
+
+    def fetch(key)
+      cache_backend.fetch(keys(key).sample)
+    end
+
+    def fetch_multi(*keys)
+      result = {}
+      cache_backend.fetch_multi(*keys.map { |key| keys(key).sample }).each do |k, v|
+        result[k[0...-2]] = v
+      end
+      result
+    end
+
+    def clear
+      cache_backend.clear
+    end
+
+    private
+
+    attr_reader(:cache_backend, :blob_replication_factor)
+
+    def keys(key)
+      blob_replication_factor.times.map { |i| "#{key}:#{i}" }
+    end
+  end
+end
--- a/test/replicated_cache_proxy_test.rb
+++ b/test/replicated_cache_proxy_test.rb
@@ -1,0 +1,43 @@
+require "test_helper"
+
+class ReplicatedCacheProxyTest < IdentityCache::TestCase
+  def setup
+    super
+    @backend = ActiveSupport::Cache::MemoryStore.new
+    @proxy = IdentityCache::ReplicatedCacheProxy.new(@backend, blob_replication_factor: 2)
+  end
+
+  def test_write_writes_both_keys
+    @proxy.write("k", "tacocat")
+    assert_equal("tacocat", @backend.fetch("k:0"))
+    assert_equal("tacocat", @backend.fetch("k:1"))
+  end
+
+  def test_deletes_both_keys
+    @backend.write("k:0", "tacocat")
+    @backend.write("k:1", "tacocat")
+    @proxy.delete("k")
+    assert_nil(@backend.fetch("k:0"))
+    assert_nil(@backend.fetch("k:1"))
+  end
+
+  def test_fetch_will_return_value_from_one_of_the_keys
+    @backend.write("k:0", "tacocat")
+    @backend.write("k:1", "tacocat")
+    assert_equal("tacocat", @proxy.fetch("k"))
+  end
+
+  def test_fetch_multi_will_return_a_mix_of_values
+    @backend.write("k:0", "tacocat")
+    @backend.write("k:1", "tacocat")
+    @backend.write("l:0", "tacodog")
+    @backend.write("l:1", "tacodog")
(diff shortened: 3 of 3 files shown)


simon-shopify:
> I'd also make the default value of 1 use the existing key value to ease in rolling this out (or maybe have the 0 index use a key without any suffix).

Good idea.

> This is kind of what I had in mind, but I'd like to be able to override the replication factor per model.

This one will need a bit of work because the cache backend is completely oblivious of the model. I'll try to come up with something that doesn't need refactoring.

lib/identity_cache/replicated_cache_proxy.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module IdentityCache
+  class ReplicatedCacheProxy
```
sirupsen: I like this as an implementation in a proxy—this is pretty clean.

lib/identity_cache/replicated_cache_proxy.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    def fetch(key)
+      cache_backend.fetch(keys(key).sample)
+    end
+
+    def fetch_multi(*keys)
+      result = {}
+      cache_backend.fetch_multi(*keys.map { |key| keys(key).sample }).each do |k, v|
+        result[k[0...-2]] = v
```
sirupsen: What's this `[0..-2]`?

sirupsen: What's this `[0..-2]`?

lib/identity_cache/replicated_cache_proxy.rb
sirupsen: Should we have a lower expiration on these than the default `6.hours` in IDC? It seems like having separate cache entries out of sync would be 😬 , at least currently you won't flip-flop. Perhaps it's fine, especially if we can already override it per model.

lib/identity_cache.rb
```diff
@@ -55,6 +56,9 @@ class << self
     mattr_accessor :fetch_read_only_records
     self.fetch_read_only_records = true
 
+    mattr_accessor :blob_replication_factor
+    self.blob_replication_factor = 1
```
sirupsen: I'm wondering if we should be able to do this per-call, and not just per-model? Do we actually see ourselves _not_ turning this on everywhere? On any path that uses IDC, shouldn't we exactly spread it across multiple servers?

test/replicated_cache_proxy_test.rb
```diff
@@ -0,0 +1,43 @@
...
+    @backend.write("k:1", "tacocat")
+    @backend.write("l:0", "tacodog")
+    @backend.write("l:1", "tacodog")
+    assert_equal({ "k" => "tacocat", "l" => "tacodog" }, @proxy.fetch_multi("k", "l"))
+  end
+
+  def test_clear_clears_the_backend
+    @backend.write("k:0", "tacocat")
+    @proxy.clear
+    assert_nil(@backend.fetch("k:0"))
```
sirupsen: Perhaps also check that `k:1` is gone

lib/identity_cache.rb
```diff
@@ -55,6 +56,9 @@ class << self
     mattr_accessor :fetch_read_only_records
     self.fetch_read_only_records = true
 
+    mattr_accessor :blob_replication_factor
```
sirupsen: We'll need to document this feature as well.

simon-shopify:
> This is awesome. I definitely think we need to figure out how we intend to use this... is there really any reason to not always turn this on for us at least? Do we really need this per model? Running it per model is basically saying "we want this on the highest throughput paths, but not on the low throughput paths".. if we need it on all high throughput paths, why does it matter to have it on low throughput paths as well? That should be much less of a performance concern in the grand scheme of things.

The more I think about this, the more I feel we should turn it on everywhere (if you enable the feature) like you suggested. One of the assumption of IDC is that the volume of reads is much higher than writes anyways. If you agree I'll start by making an application wide setting in this PR. We can still add an override on a per-model basis in the future it that is necessary.

> I'd also make the default value of 1 use the existing key value to ease in rolling this out (or maybe have the 0 index use a key without any suffix).

This won't be enough, there is still a race if a new version writes, then an old version writes on top of it during rollout. I think we need a 2-phase rollout

* Enable replicated writes.
* Wait 6 hours or whatever is the expiration time.
* Enable replicated reads.

I feel like this should be built-in in IDC, but we used monkey patches for this in the past, so it's up to you.

lib/identity_cache/replicated_cache_proxy.rb
```diff
@@ -0,0 +1,46 @@
...
+      result
+    end
+
+    def clear
+      cache_backend.clear
+    end
+
+    private
+
+    attr_reader(:cache_backend, :blob_replication_factor)
```
rafaelfranca: All `attr_reader` in this  project don't have parentheses on it so maybe it is better to keep the code consistent?

lib/identity_cache.rb
csfrancis: I would just call it `:replication_factor`.

csfrancis:
> If you agree I'll start by making an application wide setting in this PR.

👍 

> I feel like this should be built-in in IDC

I think you could do this in IDC by having a `use_read_replication_after` attribute that specifies a date. By default it could be `nil` which would use read replication right away if the `replication_factor` is > 1, but would otherwise only replicate writes until after `use_read_replication_after`.

simon-shopify:
> I think you could do this in IDC by having a use_read_replication_after attribute that specifies a date. By default it could be nil which would use read replication right away if the replication_factor is > 1, but would otherwise only replicate writes until after use_read_replication_after.

Yes, as long as you are careful with timezones (😬) that is a nice way to avoid to deploy 2 times to roll it out.

sirupsen:
> If you agree I'll start by making an application wide setting in this PR. We can still add an override on a per-model basis in the future it that is necessary.

Yeah, I agree.

lib/identity_cache.rb
```diff
@@ -72,12 +76,12 @@ def cache_backend=(cache_adaptor)
       if defined?(@cache)
         cache.cache_backend = cache_adaptor
       else
-        @cache = MemoizedCacheProxy.new(cache_adaptor)
+        @cache = wrap_in_replicated_proxy(MemoizedCacheProxy.new(cache_adaptor))
```
fbogsany: Why wrap the MemoizedCacheProxy instead of the cache backend?

lib/identity_cache/replicated_cache_proxy.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+    end
+
+    def fetch(key)
+      cache_backend.fetch(keys(key).sample)
+    end
+
+    def fetch_multi(*keys)
+      result = {}
+      cache_backend.fetch_multi(*keys.map { |key| keys(key).sample }).each do |k, v|
```
dylanahsmith: I think for the fetch methods we might want to optimize this usage to avoid allocating so many objects by creating an array of replicated keys for each requested key.  Instead, we could have a `random_replica_key(key)` that could avoid the unnecessary object allocations.

lib/identity_cache.rb
dylanahsmith: Having MemoizedCacheProxy as the backend for the replicated proxy will mean that we will store duplicate keys in memory for all the replicas.  We want to reverse the wrapping so that we memoize the requested rather than replicated key to avoid extra memory use.

On the other hand, we don't want it as the cache backend, since we want a consistent interface to work with rather than having to worry about CacheFetcher and FallbackFetcher using this proxy differently.  As such, I think we should move the cache backend wrapping logic out of MemoizedCacheProxy so that we can add a layer between it and the CacheFetcher/FallbackFetcher.

It would also be nice to allow the user to control these proxies since that ability would allow what we are doing in this PR to be implemented outside of IdentityCache itself.  It would also allow us to add rollout code using a proxy layer that does double deleting while avoiding monkey patching.

lib/identity_cache/replicated_cache_proxy.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    def write(key, value)
+      keys(key).each do |key|
+        cache_backend.write(key, value)
+      end
+    end
+
+    def delete(key)
+      keys(key).each do |key|
```
dylanahsmith: I think the code becomes harder to read by having "key" meaning two different things.  We could rename the block argument to `replica_key` to distinguish between the two types of keys.  Also, we could have an `each_replica_key` method that yields the keys without unnecessary creating an array.

  E.g.

```ruby
    def delete(key, value)
      each_replica_key(key) do |replica_key|
```

lib/identity_cache/replicated_cache_proxy.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module IdentityCache
+  class ReplicatedCacheProxy
+    def initialize(cache_backend, blob_replication_factor:)
+      @cache_backend = cache_backend
+      @blob_replication_factor = blob_replication_factor
+    end
+
+    def write(key, value)
```
dylanahsmith: I don't think `IdentityCache.cache.write` is ever used, so maybe we should just remove that and avoid having to support at each proxy layer (i.e. we could remove the write method from MemoizedCacheProxy, CacheFetcher and FallbackFetcher).

dylanahsmith:
> Wait 6 hours or whatever is the expiration time.

Although you would want the cache expiry deployed before the cache writes, you wouldn't need to wait for all the cache keys to expire in between the deploys.  That would only be necessary if you are worried about the cache miss rate on the second deploy from the replicated cache keys not being populated.

simon-shopify:
I don't think that I'm going to merge this, as it is pretty hacky and does not address the root cause of the slowness.

sirupsen:
> I don't think that I'm going to merge this, as it is pretty hacky and does not address the root cause of the slowness.

Why doesn't this address the root cause?

PR #354, opened 2017-08-11
Type cast values using attribute types before using in cache key

### LINKED ISSUE #352: Coercion missing on values passed to fetch_by_* methods
From https://github.com/Shopify/identity_cache/issues/350#issuecomment-318113933

> It looks like our cache key generation for cache index methods (i.e. fetch_by_*) just use `values.join('/')` on the raw values, which can result in different values to map to the same cache key.  This is bad because a `VendorSetting.fetch_by_languagecode('')` and `VendorSetting.fetch_by_languagecode(nil)` would use the same key, causing the former cache a result used by the latter even though they would result in a different results, since Array#join calls `to_s` on the values and `nil.to_s` becomes `''`.

> Unfortunately, we may also be relying on this behaviour in a subtle way when active record would coerce the value before performing the query.  The problem is that by changing `values.join('/')` to something like `values.map(&:inspect).join('/')` to avoid caching results from one query then using them for another query, then we could end up not expiring a cache key when a row changes if the values were coerced by active record.  For example, if `Product.fetch_by_vendor(:foo)` used `:foo` in the cache key for the provided value then active record coerces the symbol to a string based on the column type, then the cache key wouldn't be invalidated when the row is updated, since it would instead use `"foo"` in the cache key during cache invalidation.  We should be coercing the provided values based on their column type before creating the cache key.

dylanahsmith:
Fixes #352 
~~Depends on #355~~

## Problem

If we don't coerce values before using them as part of cache keys, then we can end up caching the data in a key that won't get expired when the data changes.

We also need to make sure we don't encode the values in a way that can result in conflicts between sets of values.  For example, when we just use `to_s` to encode a value in a cache key, then there can be conflicts between `nil` and `""`.

## Solution

Use `type_for_attribute(field_name).cast(value)` on the model to use active record's attribute casting logic to normalize values before using them in the cache key.

I also changed the way values are encoded in cache keys for `cache_attribute` from implicitly using `.to_s` to instead use `to_json` so that `nil` and `""` get encoded differently.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 #### 1.0.0 Unreleased
 
+- Type cast values using attribute types before using in cache key (#354)
 - Remove support for rails 4.2 (#355)
 - Set inverse cached association for cache_has_one on cache hit (#345)
 - Lazy load associated classes (#306)
--- a/lib/identity_cache/cache_key_generation.rb
+++ b/lib/identity_cache/cache_key_generation.rb
@@ -51,7 +51,7 @@ module IdentityCache
 
       private
       def rails_cache_string_for_fields_and_values(fields, values)
-        "#{fields.join('/')}:#{IdentityCache.memcache_hash(values.join('/'))}"
+        "#{fields.join('/')}:#{IdentityCache.memcache_hash(values.map(&:to_json).join('/'))}"
       end
     end
 
--- a/lib/identity_cache/cache_key_generation.rb
+++ b/lib/identity_cache/cache_key_generation.rb
@@ -51,7 +51,7 @@ module IdentityCache
 
       private
       def rails_cache_string_for_fields_and_values(fields, values)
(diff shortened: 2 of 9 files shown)


dylanahsmith:
Looks like we also need to either remove support for rails 4.2 or have version conditional code, since the `cast` method was called something different in that version of rails.  The next release will have breaking changes, so we may as well just remove support for rails 4.2.

dylanahsmith:
> In that case, it's not unreasonable to have a switch case that adds something like s for strings and n for nils to remove the ambiguity.

The strings still need to be quoted to avoid ambiguity.  E.g. a cache index with two fields would have the same cache key for the values `["http://example.com", "foo/bar"]` and `["http://example.com/foo", "bar"]`, both of which would encode the joined values as `http://example.com/foo/bar` right now.  However, maybe I could just encode values by using `value.try!(:to_s).inspect` so that it both doesn't convert `nil` to an `""` and also quotes strings to avoid cache key conflicts between two strings.  Since we are already using `.to_s` to encode values, then that should be safe for backwards compatibility.

PR #360, opened 2017-11-15
Set inverse asoociation for cache_has_many without :embed => true

### LINKED ISSUE #357: should inverse_of work?
Not sure if it's a misconfiguration on my side or if it's simply not supported. Given:

```
class Person < ActiveRecord::Base
  include IdentityCache
  belongs_to :client, inverse_of: :persons
  cache_belongs_to :client
end

class Client < ActiveRecord::Base
  include IdentityCache
  has_many :persons, -> { where partner_id: nil }, dependent: :destroy, inverse_of: :client
  cache_has_many :persons
end
```

Without the cache the client isn't fetched again when we fetch persons on a client:
```
irb(main):002:0> client = Client.find(4)
  Client Load (0.5ms)  SELECT  `clients`.* FROM `clients` WHERE `clients`.`id` = 4 LIMIT 1
=> #<Client id: 4, name: .... >
irb(main):003:0> client.persons.map{|p| p.client.object_id}
  Person Load (0.7ms)  SELECT `persons`.* FROM `persons` WHERE `persons`.`deleted_at` IS NULL AND `persons`.`client_id` = 4 AND `persons`.`partner_id` IS NULL
=> [70220559135500, 70220559135500, 70220559135500, 70220559135500, 70220559135500, 70220559135500, 70220559135500, 70220559135500]
```

With the IdentityCache I get a client query for every person and different objects. Shouldn't IdentityCache simply setup the association to the existing client object?
```
irb(main):002:0> client = Client.find(4)
  Client Load (0.5ms)  SELECT  `clients`.* FROM `clients` WHERE `clients`.`id` = 4 LIMIT 1
=> #<Client id: 4, name: .... >
irb(main):002:0> client.fetch_persons.map{|p| p.client.object_id}
   (1.2ms)  SELECT `persons`.`id` FROM `persons` WHERE `persons`.`deleted_at` IS NULL AND `persons`.`client_id` = 4 AND `persons`.`partner_id` IS NULL
Cache cas_multi: greenlight:1:IDC:7:blob:Person:7919784493729940308:9/IDC:7:blob:Person:7919784493729940308:10/IDC:7:blob:Person:7919784493729940308:11/IDC:7:blob:Person:7919784493729940308:3459/IDC:7:blob:Person:79197844937299403:md5:1e1faf6c2624ebf80c00411dbf35dcc9 ({:namespace=>"greenlight:1", :expires_in=>21600})
  Person Load (0.8ms)  SELECT `persons`.* FROM `persons` WHERE `persons`.`deleted_at` IS NULL AND `persons`.`id` IN (9, 10, 11, 3459, 4135, 4147, 4312, 4458)
Cache write: greenlight:1:IDC:7:blob:Person:7919784493729940308:9 ({:namespace=>"greenlight:1", :expires_in=>21600, :unless_exist=>true})
Cache write: greenlight:1:IDC:7:blob:Person:7919784493729940308:10 ({:namespace=>"greenlight:1", :expires_in=>21600, :unless_exist=>true})
Cache write: greenlight:1:IDC:7:blob:Person:7919784493729940308:11 ({:namespace=>"greenlight:1", :expires_in=>21600, :unless_exist=>true})
Cache write: greenlight:1:IDC:7:blob:Person:7919784493729940308:3459 ({:namespace=>"greenlight:1", :expires_in=>21600, :unless_exist=>true})
Cache write: greenlight:1:IDC:7:blob:Person:7919784493729940308:4135 ({:namespace=>"greenlight:1", :expires_in=>21600, :unless_exist=>true})
Cache write: greenlight:1:IDC:7:blob:Person:7919784493729940308:4147 ({:namespace=>"greenlight:1", :expires_in=>21600, :unless_exist=>true})
Cache write: greenlight:1:IDC:7:blob:Person:7919784493729940308:4312 ({:namespace=>"greenlight:1", :expires_in=>21600, :unless_exist=>true})
Cache write: greenlight:1:IDC:7:blob:Person:7919784493729940308:4458 ({:namespace=>"greenlight:1", :expires_in=>21600, :unless_exist=>true})
[IdentityCache] cache miss for IDC:7:blob:Person:7919784493729940308:9 (multi)
[IdentityCache] cache miss for IDC:7:blob:Person:7919784493729940308:10 (multi)
[IdentityCache] cache miss for IDC:7:blob:Person:7919784493729940308:11 (multi)
[IdentityCache] cache miss for IDC:7:blob:Person:7919784493729940308:3459 (multi)
[IdentityCache] cache miss for IDC:7:blob:Person:7919784493729940308:4135 (multi)
[IdentityCache] cache miss for IDC:7:blob:Person:7919784493729940308:4147 (multi)
[IdentityCache] cache miss for IDC:7:blob:Person:7919784493729940308:4312 (multi)
[IdentityCache] cache miss for IDC:7:blob:Person:7919784493729940308:4458 (multi)
  Client Load (0.6ms)  SELECT  `clients`.* FROM `clients` WHERE `clients`.`id` = 4 LIMIT 1
  Client Load (0.5ms)  SELECT  `clients`.* FROM `clients` WHERE `clients`.`id` = 4 LIMIT 1
  Client Load (0.6ms)  SELECT  `clients`.* FROM `clients` WHERE `clients`.`id` = 4 LIMIT 1
  Client Load (0.5ms)  SELECT  `clients`.* FROM `clients` WHERE `clients`.`id` = 4 LIMIT 1
  Client Load (0.6ms)  SELECT  `clients`.* FROM `clients` WHERE `clients`.`id` = 4 LIMIT 1
  Client Load (0.5ms)  SELECT  `clients`.* FROM `clients` WHERE `clients`.`id` = 4 LIMIT 1
  Client Load (0.5ms)  SELECT  `clients`.* FROM `clients` WHERE `clients`.`id` = 4 LIMIT 1
  Client Load (0.5ms)  SELECT  `clients`.* FROM `clients` WHERE `clients`.`id` = 4 LIMIT 1
=> [70220565335420, 70220557688780, 70220557994400, 70220558006100, 70220589297320, 70220589903960, 70220589965920, 70220560291960]
```

roccoblues:
Fixes #357

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Raise when trying to cache a belong_to association with a scope. Previously the scope was ignored on a cache hit (#323)
 - Remove deprecated `never_set_inverse_association` option (#319)
 - Fix column name in the preload association query when using custom primary keys (#338)
+- Fix inverse asoociation setting for cache_has_many without `:embed => true`
 
 #### 0.5.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 #### 1.0.0 Unreleased
 
+- Set inverse asoociation for cache_has_many without `:embed => true`
 - Type cast values using attribute types before using in cache key (#354)
 - Remove support for rails 4.2 (#355)
 - Set inverse cached association for cache_has_one on cache hit (#345)
@@ -11,7 +12,6 @@
 - Raise when trying to cache a belong_to association with a scope. Previously the scope was ignored on a cache hit (#323)
 - Remove deprecated `never_set_inverse_association` option (#319)
 - Fix column name in the preload association query when using custom primary keys (#338)
-- Fix inverse asoociation setting for cache_has_many without `:embed => true`
 
 #### 0.5.1
 
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -217,6 +217,8 @@ module IdentityCache
         options[:records_variable_name]   = "cached_#{association}"
         options[:prepopulate_method_name] = "prepopulate_fetched_#{association}"
 
+        inverse_name = options[:association_reflection].inverse_of.try(:name)
+
         self.class_eval(<<-CODE, __FILE__, __LINE__ + 1)
           attr_reader :#{options[:ids_variable_name]}
 
@@ -227,7 +229,12 @@ module IdentityCache
           def #{options[:cached_accessor_name]}
             association_klass = association(:#{association}).klass
             if association_klass.should_use_cache? && !#{association}.loaded?
-              @#{options[:records_variable_name]} ||= #{options[:association_reflection].class_name}.fetch_multi(#{options[:cached_ids_name]})
+              @#{options[:records_variable_name]} ||= begin
+                #{options[:association_reflection].class_name}.fetch_multi(#{options[:cached_ids_name]}).map do |record|
+                  record.send('#{inverse_name}=', self) if #{inverse_name.present?}
+                  record
+                end
+              end
             else
               #{association}.to_a
             end
--- a/test/inverse_of_test.rb
+++ b/test/inverse_of_test.rb
@@ -1,0 +1,63 @@
+require "test_helper"
+
+class PrefetchAssociationsTest < IdentityCache::TestCase
+  def setup
+    super
+    @item   = Item.create!(:title => 'bob')
+
+    AssociatedRecord.create!(:item => @item)
+    AssociatedRecord.create!(:item => @item)
+    AssociatedRecord.create!(:item => @item)
+  end
+
+  def test_fetch_should_setup_association_for_cache_has_one
+    Item.cache_has_one :associated
+
+    item = Item.find_by_id(@item.id)
+
+    record = item.fetch_associated
+    assert_equal item.object_id, record.item.object_id
+  end
+
+  def test_fetch_should_setup_association_for_cache_has_one_embedded
+    Item.cache_has_one :associated, :embed => true
+
+    item = Item.find_by_id(@item.id)
+
+    record = item.fetch_associated
+    assert_equal item.object_id, record.item.object_id
+  end
+
+  def test_fetch_should_setup_association_for_cache_has_many
+    Item.cache_has_many :associated_records
+
+    item = Item.find_by_id(@item.id)
+
+    records = item.fetch_associated_records
+    records.each do |r|
+      assert_equal item.object_id, r.item.object_id
+    end
+  end
+
+  def test_fetch_should_setup_association_for_cache_has_many_embedded
+    Item.cache_has_many :associated_records, :embed => true
+
+    item = Item.find_by_id(@item.id)
+
+    records = item.fetch_associated_records
+    records.each do |r|
+      assert_equal item.object_id, r.item.object_id
+    end
+  end
+
+  def test_fetch_should_setup_association_for_cache_has_many_embedded_ids
+    Item.cache_has_many :associated_records, :embed => :ids
+
+    item = Item.find_by_id(@item.id)
+
+    records = item.fetch_associated_records
+    records.each do |r|
+      assert_equal item.object_id, r.item.object_id
+    end
+  end
+end
--- a/test/inverse_of_test.rb
+++ b/test/inverse_of_test.rb
@@ -1,17 +1,18 @@
 require "test_helper"
 
-class PrefetchAssociationsTest < IdentityCache::TestCase
+class InverseOfTest < IdentityCache::TestCase
   def setup
     super
-    @item   = Item.create!(:title => 'bob')
+
+    @item = Item.create!(:title => 'bob')
 
     AssociatedRecord.create!(:item => @item)
     AssociatedRecord.create!(:item => @item)
     AssociatedRecord.create!(:item => @item)
   end
 
-  def test_fetch_should_setup_association_for_cache_has_one
-    Item.cache_has_one :associated
+  def test_fetch_should_setup_association_for_cache_belongs_to
+    Item.cache_belongs_to :associated
 
     item = Item.find_by_id(@item.id)
 
@@ -19,8 +20,8 @@ class PrefetchAssociationsTest < IdentityCache::TestCase
     assert_equal item.object_id, record.item.object_id
   end
 
-  def test_fetch_should_setup_association_for_cache_has_one_embedded
-    Item.cache_has_one :associated, :embed => true
+  def test_fetch_should_setup_association_for_cache_has_one
+    Item.cache_has_one :associated
 
     item = Item.find_by_id(@item.id)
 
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 #### 1.0.0 Unreleased
 
+- Set inverse asoociation for cache_has_many without `:embed => true`
```
shwetabhgarg: asoociation -> association :P

lib/identity_cache/configuration_dsl.rb
```diff
@@ -217,6 +217,8 @@ def build_id_embedded_has_many_cache(association, options) #:nodoc:
         options[:records_variable_name]   = "cached_#{association}"
         options[:prepopulate_method_name] = "prepopulate_fetched_#{association}"
 
+        inverse_name = options[:association_reflection].inverse_of.try(:name)
```
dylanahsmith: Prefer `try!` over `try`, since the latter also does a respond_to? check that can hide bugs.

lib/identity_cache/configuration_dsl.rb
```diff
@@ -227,7 +229,12 @@ def #{options[:cached_ids_name]}
           def #{options[:cached_accessor_name]}
             association_klass = association(:#{association}).klass
             if association_klass.should_use_cache? && !#{association}.loaded?
-              @#{options[:records_variable_name]} ||= #{options[:association_reflection].class_name}.fetch_multi(#{options[:cached_ids_name]})
+              @#{options[:records_variable_name]} ||= begin
+                #{options[:association_reflection].class_name}.fetch_multi(#{options[:cached_ids_name]}).map do |record|
+                  record.send('#{inverse_name}=', self) if #{inverse_name.present?}
```
dylanahsmith: We don't want to set the association itself, since putting records loaded from the cache into an active record association would make it easier to unintentionally persist that data to the database, which could persist cache corruption.  Instead, we want to make sure they use the cached accessor to access it.  Can you re-use `set_inverse_of_cached_association` to do it the same way as embedded cached associations?

test/inverse_of_test.rb
```diff
@@ -0,0 +1,53 @@
...
+    Item.cache_has_one :associated
+
+    item = Item.find_by_id(@item.id)
+
+    record = item.fetch_associated
+    assert_equal item.object_id, record.item.object_id
+  end
+
+  def test_fetch_should_setup_association_for_cache_has_many
+    Item.cache_has_many :associated_records
```
dylanahsmith: Explicitly use the `:embed => :ids` option for clarity

test/inverse_of_test.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    item = Item.find_by_id(@item.id)
+
+    records = item.fetch_associated_records
+    records.each do |r|
+      assert_equal item.object_id, r.item.object_id
```
dylanahsmith: You should need to use `r.fetch_item`.

test/inverse_of_test.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+  end
+
+  def test_fetch_should_setup_association_for_cache_has_many_embedded
+    Item.cache_has_many :associated_records, :embed => true
+
+    item = Item.find_by_id(@item.id)
+
+    records = item.fetch_associated_records
```
dylanahsmith: This will always just delegate to the active record association, since `item` wasn't fetched from the cache.

test/inverse_of_test.rb
```diff
@@ -0,0 +1,53 @@
+require "test_helper"
+
+class InverseOfTest < IdentityCache::TestCase
```
dylanahsmith: Aren't there already tests for setting the inverse cached association?  If there are already tests in denormalized_has_many_test.rb, denormalized_has_one_test.rb and normalized_belongs_to_test.rb then we just need tests for this behaviour in normalized_has_many_test.rb

test/inverse_of_test.rb
```diff
@@ -0,0 +1,53 @@
...
+    assert_equal item.object_id, record.item.object_id
+  end
+
+  def test_fetch_should_setup_association_for_cache_has_one
+    Item.cache_has_one :associated
+
+    item = Item.find_by_id(@item.id)
+
+    record = item.fetch_associated
+    assert_equal item.object_id, record.item.object_id
```
dylanahsmith: This will always just delegate to the active record association, since item wasn't fetched from the cache and this is an embedded cached association.

test/inverse_of_test.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    @item = Item.create!(:title => 'bob')
+
+    AssociatedRecord.create!(:item => @item)
+    AssociatedRecord.create!(:item => @item)
+    AssociatedRecord.create!(:item => @item)
+  end
+
+  def test_fetch_should_setup_association_for_cache_belongs_to
+    Item.cache_belongs_to :associated
```
dylanahsmith: This seems like it should raise, since the model doesn't even have a `belongs_to :associated` association, nor does the table have an associated_id column on the items table.  I think this must be actually using the has_one association.

CHANGELOG.md
dylanahsmith: I would say it is for an "id embedded cach_has_many associations" rather than saying what it isn't for.

test/inverse_of_test.rb
```diff
@@ -0,0 +1,53 @@
...
+    item = Item.find_by_id(@item.id)
+
+    record = item.fetch_associated
+    assert_equal item.object_id, record.item.object_id
+  end
+
+  def test_fetch_should_setup_association_for_cache_has_many
+    Item.cache_has_many :associated_records
+
+    item = Item.find_by_id(@item.id)
```
dylanahsmith: Use `find` when you always expect it to return a record.

test/inverse_of_test.rb
roccoblues: No really, only this: https://github.com/Shopify/identity_cache/blob/master/test/denormalized_has_many_test.rb#L105 and I'm not sure what it tests.

roccoblues:
@dylanahsmith I have the feeling we've been talking about different things. Can you have a look at the code in the related issue again? Here's what I meant in pseudo-code:

```
item = Item.find(ID) # fetches item from the db
associated_records = item.fetch_associated # fetches from cache

# now I wanted the item accessor on the records fetched from cache
# to *not* load the item again
associated_records.first.item.object_id == item.object_id
```

I just found that you even have a test that this should explicitly not happen: https://github.com/Shopify/identity_cache/blob/master/test/denormalized_has_many_test.rb#L124 ?

Did you instead think about this?
```
item = Item.fetch(ID) # fetches item from the cache
associated_records = item.fetch_associated # fetches from cache

# the fetch_ method should return the already fetch item from cache?
associated_records.first.fetch_item.object_id == item.object_id
```

dylanahsmith:
> I just found that you even have a test that this should explicitly not happen: https://github.com/Shopify/identity_cache/blob/master/test/denormalized_has_many_test.rb#L124 ?

In that case it is using the active record association `associated_record.item`, instead of the cache accessor `associated_record.fetch_item`, so it shouldn't need modification.  However, you could also add an assertion of `assert_equal associated_record.fetch_item.object_id, item.object_id` which should also pass with your change.  The difference is that we don't want to store the record loaded from the cache in the active record association itself, we want to store it in the instance variable that the `fetch_#{association}` method accesses.

roccoblues:
Ok, I think I got it now. Please check again.

PR #363, opened 2018-01-25
Fix typo in README sample

dlackty:


dylanahsmith:
Thanks for clearing that up!

PR #365, opened 2018-03-14
Handle missing constants as misses

denislins:
While caching embedded associations, IDC caches the model names (aka constants) inside the cache blob. However, when fetching the data a constant might not exist anymore, i.e. it might have been moved, or deleted. This currently raises a `NameError`.

This PR aims to fix this issue by handling these cases as normal misses, hitting the database to fetch the requested data. This also deletes the key from the cache since it's no longer valid.

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -24,13 +24,19 @@ module IdentityCache
         return unless id
         record = if should_use_cache?
           require_if_necessary do
-            object = nil
-            coder = IdentityCache.fetch(rails_cache_key(id)){ instrumented_coder_from_record(object = resolve_cache_miss(id)) }
-            object ||= instrumented_record_from_coder(coder)
-            if object && object.id != id
-              IdentityCache.logger.error "[IDC id mismatch] fetch_by_id_requested=#{id} fetch_by_id_got=#{object.id} for #{object.inspect[(0..100)]}"
+            cache_key = rails_cache_key(id)
+            begin
+              object = nil
+              coder = IdentityCache.fetch(cache_key){ instrumented_coder_from_record(object = resolve_cache_miss(id)) }
+              object ||= instrumented_record_from_coder(coder)
+              if object && object.id != id
+                IdentityCache.logger.error "[IDC id mismatch] fetch_by_id_requested=#{id} fetch_by_id_got=#{object.id} for #{object.inspect[(0..100)]}"
+              end
+              object
+            rescue InexistentAssociationError
+              IdentityCache.cache.delete(cache_key)
+              resolve_cache_miss(id)
             end
-            object
           end
         else
           resolve_cache_miss(id)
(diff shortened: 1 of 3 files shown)


lib/identity_cache/query_api.rb
```diff
@@ -132,7 +143,12 @@ def instrumented_record_from_coder(coder) #:nodoc:
 
       def record_from_coder(coder) #:nodoc:
         if coder
-          klass = coder[:class].constantize
+          begin
+            klass = coder[:class].constantize
+          rescue NameError
+            raise InexistentAssociationError
+          end
```
denislins: I decided to use this constant to avoid conflicts with other similar errors. This can be useful when developing the lib, as a missing class would be handled as a miss and would be way harder to find the issue.

camilo: hmm, it is reasonable to specify a class in this context for sure, what I'm wondering here is about the inheritance situation, why is`class InexistentAssociationError < StandardError; end` not a `NameError`  subclass? I could make an OOP argument for that, but also there is the more pragmatic we probably still want the exception when bubbling up to tell us what constant it was looking for, for ease of debugging.

WDYT?

lib/identity_cache/query_api.rb
```diff
@@ -24,13 +24,19 @@ def fetch_by_id(id, includes: nil)
         return unless id
         record = if should_use_cache?
           require_if_necessary do
-            object = nil
-            coder = IdentityCache.fetch(rails_cache_key(id)){ instrumented_coder_from_record(object = resolve_cache_miss(id)) }
-            object ||= instrumented_record_from_coder(coder)
-            if object && object.id != id
-              IdentityCache.logger.error "[IDC id mismatch] fetch_by_id_requested=#{id} fetch_by_id_got=#{object.id} for #{object.inspect[(0..100)]}"
+            cache_key = rails_cache_key(id)
+            begin
```
camilo: I find both `fetch_by_id` and the multi version are getting too chunky, is there a way to factor out the new logic into a function that we can call here ?

camilo:
I'm fine with the approach for now, we might want to take a hard look at how class/model names are serialized in the future, one think I dislike is the fact that the logic for embedded vs non-embeded models is divergent. A purely speculative comment would be, can we somehow add ALL the constant names to the key?

dylanahsmith:
I would prefer to not depend on the class name that is in the cache at all, so I've opened up an alternative PR: https://github.com/Shopify/identity_cache/pull/366

Also, I would prefer if the tests didn't rely as much on the internals, so the implementation can be refactored without updating a bunch of tests with mocked data or hard-coded keys.  As such, I wasn't able to re-use the tests from this PR in my alternative PR.

lib/identity_cache/query_api.rb
denislins: Makes complete sense!

PR #366, opened 2018-03-15
Fix exceptions from renaming classes by not using a cached name

dylanahsmith:
Closes #365

## Problem

We were storing the cache name for records in the cache, so renaming a class would result in a NameError when using cache entries from before the rename.

## Solution

We don't actually need the class name, since `fetch_by_id` and `fetch_multi` are called on the class and associated classes we can get from the association.

The class name is still stored in the class for now, so this can be rolled out without causing problems during deploys or if a rollback is needed.

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -106,6 +106,22 @@ module IdentityCache
         nil
       end
 
+      protected
+
+      def record_from_coder(coder) #:nodoc:
+        if coder
+          record = instantiate(coder[:attributes].dup)
+
+          coder[:associations].each do |name, value|
+            associated_class = reflect_on_association(name).klass
+            set_embedded_association(record, name, hydrate_association_target(associated_class, value))
+          end if coder.has_key?(:associations)
+          coder[:association_ids].each {|name, ids| record.instance_variable_set(:"@#{record.class.cached_has_manys[name][:ids_variable_name]}", ids) } if coder.has_key?(:association_ids)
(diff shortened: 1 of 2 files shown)


lib/identity_cache/query_api.rb
```diff
@@ -106,6 +106,22 @@ def prefetch_associations(associations, records)
         nil
       end
 
+      protected
+
+      def record_from_coder(coder) #:nodoc:
+        if coder
+          record = instantiate(coder[:attributes].dup)
+
+          coder[:associations].each do |name, value|
```
camilo: so to be sure the `name` here is the name of the association NOT of the associated model.

lib/identity_cache/query_api.rb
```diff
@@ -106,6 +106,22 @@ def prefetch_associations(associations, records)
...
+            associated_class = reflect_on_association(name).klass
+            set_embedded_association(record, name, hydrate_association_target(associated_class, value))
+          end if coder.has_key?(:associations)
```
camilo: style nit: I'd go for more nesting

```

if coder.has_key?(...)
  loop here
end

```

take it or leave it

lib/identity_cache/query_api.rb
dylanahsmith: Correct.  That is why `reflect_on_association(name)` on the following line works

lib/identity_cache/query_api.rb
dylanahsmith: I hadn't touched this line, other than making the method protected, but sure.

PR #367, opened 2018-04-26
[WIP] Implement Dalli cache fetcher

maximebedard:
@fbogsany Currently a huge WIP, but any feedback is appreciated 😄


rafaelfranca:
Any reason why using dali cache store? Active Support memcache store uses Dalli as a client and have better compatibility with Rails.

rafaelfranca:
We can definitely upstream cas support, if dalli has support and our custom store also does it may be useful for everyone

maximebedard:
The problem I have with merge this upstream is that it's not a interface that works for all of the Cache::Store. Which somewhat defeats the purpose of the interface a little bit?

rafaelfranca:
There is no idea of interface. Yes, there is a super class but not every adapter needs to implement all the same features. For example, the file and memory adapter don't have the `_muilti` methods. The documentation of the `Store` super class even mention that each implementation can have specific features https://github.com/rails/rails/blob/8d4e0ab33d5aa4ad46bc648a0636aa788df828cb/activesupport/lib/active_support/cache.rb#L115-L116

maximebedard:
Cool thanks for the feedback, TIL. I'll make sure everything works here and in core, then open a PR in rails to add support for cas/cas_multi and document them appropriately.

maximebedard:
Going to close this one and try to open a PR upstream shortly.

PR #370, opened 2018-07-04
Log delete failed on error level

Nucc:
Would you mind if I changed the log severity to `error` when invalidating a record failed?

When `expire_cache` fails because of a network blip on the production system and Memcached becomes unreachable for a moment seems to be a bigger issue to log it on debug level. Production systems use info level as default and if invalidating the cache fails then the system can go to an inconsistent state easily when Memcached comes back, that's why I think this message should be on error level instead.


dylanahsmith:
Thanks.  I bet that will come in handy when debugging the effects of memcache connection issues in production.

PR #372, opened 2018-09-06
Delay hydration of cache embedded associations until accessed

dylanahsmith:
Depends on #371

## Problem

The hydration instrumentation added in #340 showed that we spend a significant amount of time for the hydration of cache blobs with a lot of embedded associations.  We should try to improve performance around hydration since we use identity cache and embedded association so heavily.

## Solution

I don't think we even need the embedded associations in many cases, so I think we can just lazily hydrate embedded associations.

Unfortunately, doing this with the current instrumentation would cause a lot more instrumentation overhead, since we wouldn't be able to group all the hydration together to instrument it once for all embedded associations.  Even if we kept the instrumentation, it would be hard to compare with the previous instrumentation data.  As a result, I decided to remove the hydration instrumentation as part of this PR.

### DIFF
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -191,11 +191,16 @@ module IdentityCache
         options[:association_reflection] = reflect_on_association(association)
         options[:cached_accessor_name]   = "fetch_#{association}"
         options[:records_variable_name]  = :"@cached_#{association}"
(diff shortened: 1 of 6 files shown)


dylanahsmith:
If we want to instrument this in the future, then I think we should keep track of the hydration across the unit of work.  We could keep track of the number of records hydrated and the total time spent doing the hydration.

PR #379, opened 2019-01-30
Use Active Record association if already loaded on embedded associations

dylanahsmith:
Follow-up to #378

## Problem

We have a difference between ID embedded and recursively embedded associations with respect to what happens when the association is loaded from the cache, but then the active record association is loaded and possibly modified.  ID embedded associations use the active record association once it has been loaded on the cache association accessor, but recursively embedded associations continue using the possibly stale association target from the cache fetch.  Rails 6 also changes whether the association target is mutated in-place, which is relevant if the array is stored in an instance variable for the `fetch_#{association_name}` method and returned even after the association is loaded.

## Solution

Use the active record association if it has been loaded for recursively embedded associations


Edouard-chin:
Taking the liberty to merge your PR, hope you won't mind 😺

PR #380, opened 2019-02-13
Sprinkled AS::Notifications all over

alexsnaps:
Adds more support for AS::Nofications. Mostly adding back `hydration.identity_cache`, but also adds `association_fetch` notifications for prefetching...


alexsnaps:
@dylanahsmith 👋 Let me know what you think…

lib/identity_cache/query_api.rb
```diff
@@ -106,9 +106,16 @@ def prefetch_associations(associations, records)
         nil
       end
 
-      def record_from_coder(coder) #:nodoc:
-        return nil unless coder
+      def instrumented_record_from_coder(coder) #:nodoc:
+        return unless coder
```
dylanahsmith: Nitpick: I would prefer `return nil` over `return` when the return value is meant to be used.

PR #382, opened 2019-03-06
Make expire_primary_index a class method

DerekStride:
### Problem

Sometimes we want to expire the primary key index after updating the record using raw SQL is an `ActiveRecord#update_all` call. For example in the case of doing an atomic increment as seen in the snippet below.

```ruby
Model.where(id: record.id).update_all("#{column} = #{column} + 1")
```

### Solution

Make `expire_primary_index` delegate to a public class method called `expire_primary_key_cache_index(id)` and move the current logic to the new method.

**Example use case:**

```ruby
def increment(id, column)
  Model.where(id: id).update_all("#{column} = #{column} + 1")
  expire_primary_key_cache_index(id)
end
```

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -113,7 +113,8 @@ module IdentityCache
         end
       end
 
-      def expire_primary_index(id) # :nodoc:
+      # Invalidates the primary cache index for the associated record. Will not invalidate cached attributes.
+      def expire_primary_key_cache_index(id)
         return unless primary_cache_index_enabled
         id = type_for_attribute(primary_key).cast(id)
         IdentityCache.cache.delete(rails_cache_key(id))
@@ -458,6 +459,13 @@ module IdentityCache
       end
     end
 
+    # Invalidate the cache data associated with the record.
+    def expire_cache
+      expire_primary_index
+      expire_attribute_indexes
+      true
+    end
+
     private
 
     def fetch_recursively_cached_association(ivar_name, dehydrated_ivar_name, association_name) # :nodoc:
@@ -514,7 +522,7 @@ module IdentityCache
     end
 
     def expire_primary_index # :nodoc:
-      self.class.expire_primary_index(id)
+      self.class.expire_primary_key_cache_index(id)
     end
 
     def expire_attribute_indexes # :nodoc:
@@ -530,12 +538,6 @@ module IdentityCache
           end
         end
       end
-    end
-
-    def expire_cache # :nodoc:
-      expire_primary_index
-      expire_attribute_indexes
-      true
     end
 
     def was_new_record? # :nodoc:
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -113,6 +113,12 @@ module IdentityCache
         end
       end
 
+      def expire_primary_index(id) # :nodoc:
+        return unless primary_cache_index_enabled
+        id = type_for_attribute(primary_key).cast(id)
+        IdentityCache.cache.delete(rails_cache_key(id))
+      end
+
       private
 
       def record_from_coder(coder) #:nodoc:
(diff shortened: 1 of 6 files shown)


lib/identity_cache/query_api.rb
```diff
@@ -450,6 +450,24 @@ def prefetch_one_association(association, records)
...
+          extra_keys =
+            if respond_to?(:updated_at)
+              old_updated_at = old_values_for_fields([:updated_at]).first
+              "expiring_last_updated_at=#{old_updated_at}"
+            else
+              ""
+            end
+
+          "[IdentityCache] expiring=#{name} expiring_id=#{id} #{extra_keys}"
+        end
```
dylanahsmith: This debugging code depends on more than just the `id` from the instance.  I think we can just remove it.  I didn't even notice this code before today since we don't usually use the debug log level.

DerekStride:
Updated the PR to remove the debugging code like you suggested @dylanahsmith

lib/identity_cache/query_api.rb
```diff
@@ -508,21 +513,7 @@ def set_inverse_of_cached_association(association_options, association_target)
...
-            "expiring_last_updated_at=#{old_updated_at}"
-          else
-            ""
-          end
-
-        "[IdentityCache] expiring=#{self.class.name} expiring_id=#{id} #{extra_keys}"
-      end
-
-      IdentityCache.cache.delete(primary_cache_index_key)
+      self.class.send(:expire_primary_index, id)
```
dylanahsmith: Maybe we should make `expire_primary_index` and `expire_cache` public to make it clearer that we depend on them in our application, so they aren't really just internal library methods.

lib/identity_cache/query_api.rb
```diff
@@ -450,6 +450,11 @@ def prefetch_one_association(association, records)
         end
         next_level_records
       end
+
+      def expire_primary_index(id) # :nodoc:
+        return unless primary_cache_index_enabled
+        IdentityCache.cache.delete(rails_cache_key(id))
```
dylanahsmith:
```suggestion
        id = type_for_attribute(primary_key).cast(id)
        IdentityCache.cache.delete(rails_cache_key(id))
```

might make sense with this being a public method to protect against mis-use (e.g. passing in a string id)

lib/identity_cache/query_api.rb
DerekStride: I opted to keep them private so that other objects can't use them. Maybe protected could make sense here if the intent is to make it clear that they are safe to use outside of the library.

dylanahsmith: We would still need to use `self.class.send(:expire_primary_index, id)` even if the method were `protected`.  I'm not sure if makes sense to keep methods like `expire_cache` private given that active record already exposes lots of low level public methods on active record objects (e.g. `delete`, `update_column`).

DerekStride: I've made `expire_primary_index` public, I can make `expire_cache` public in this PR too if you want.

dylanahsmith: > I can make expire_cache public in this PR too if you want

Yes, please

DerekStride: Should it make `expire_attribute_indexes` and the instance method `expire_primary_index` public too? I just made `expire_cache` in my additional commit but I can move all of them.

dylanahsmith: If they have an instance, then it seems preferable for them to use `expire_cache`

lib/identity_cache/query_api.rb
```diff
@@ -113,6 +113,12 @@ def instrumented_record_from_coder(coder) #:nodoc:
         end
       end
 
+      def expire_primary_index(id) # :nodoc:
```
rafaelfranca: If we want people to use it we should remove the `nodoc` and also document what it does, probably with a better name. This name is confusing. For example, when I see this method call I think: expire the primary index of what?

DerekStride: I've updated the public method to be named `expire_primary_cache_index_by_id` and added a description.

dylanahsmith: > For example, when I see this method call I think: expire the primary index of what?

We have multiple references to this primary index, e.g.
* IdentityCache::WithoutPrimaryIndex
* primary_cache_index_enabled
* https://github.com/Shopify/identity_cache#basic-usage briefly explains that fetching by id is using the primary index

Perhaps it would be clearer if we referred to it as the primary **key** cache index for consistency with `primary_key` active record method and the concept of a primary key in the database.

> I've updated the public method to be named `expire_primary_cache_index_by_id`

Since the primary cache index is always by id, the `by_id` suffix seems kind of redundant and like there could be some primary cache index by something else.  Instead, I think `expire_primary_key_cache_index` would be a better name for this method by just clarifying the reference to the primary key.

PR #383, opened 2019-07-15
Fix cached_has_many for polymorphic associations

cbothner:
When Product and Shop both `has_many :translations, as: translatable` with `cache_has_many :translations`, `fetch_translations` returns all translations where `translatable_id` matches, without filtering for `translatable_type == 'Product'`. This patch fixes it, but maybe we're missing an obvious configuration point? Otherwise, I'm happy to add tests / iterate on the implementation.

cc: @christianblais

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -223,6 +223,7 @@ module IdentityCache
         reflection = association_options.fetch(:association_reflection)
         child_model = reflection.klass
         scope = child_model.all
+        scope = scope.where(reflection.type => base_class.name) if reflection.type
         scope = scope.instance_exec(nil, &reflection.scope) if reflection.scope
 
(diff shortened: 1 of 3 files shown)


christianblais:
@dylanahsmith the TLDR is, when IDC hydrates the cache, it currently disregards polymorphic associations (ignoring the `type` column`). Here, we're using the [`type` attribute from ActiveRecord::Reflection::AssociationReflection](https://github.com/rails/rails/blob/5a9301ce47f1117c8d457cad2b850f932bf7f518/activerecord/lib/active_record/reflection.rb#L430) to properly set the scope when needed.

dylanahsmith:
Change looks good to me, but could use a regression test.

cbothner:
@dylanahsmith any idea what's up with the errors in the CI build?

PR #388, opened 2019-09-03
Add a Railtie to automatically eager load IdentityCache in Rails context

casperisfine:
`IdentityCache` [has an `eager_load!` method](https://github.com/Shopify/identity_cache/blob/ff26ed60abef959b667cecac7265dd370041c92b/lib/identity_cache.rb#L153-L155), so it should be added to `app.config.eager_load_namespaces`.

Might as well add a Railtie so that it's done automatically.

### DIFF
--- a/lib/identity_cache.rb
+++ b/lib/identity_cache.rb
@@ -163,3 +163,5 @@ module IdentityCache
     end
   end
 end
+
+require 'identity_cache/railtie' if defined?(Rails)
--- a/lib/identity_cache/railtie.rb
+++ b/lib/identity_cache/railtie.rb
@@ -1,0 +1,7 @@
+module IdentityCache
+  class Railtie < Rails::Railtie
+    initializer "identity_cache.setup" do |app|
+      app.config.eager_load_namespaces << IdentityCache
+    end
+  end
+end


lib/identity_cache.rb
```diff
@@ -163,3 +163,5 @@ def fetch_in_batches(keys)
     end
   end
 end
+
+require 'identity_cache/railtie' if defined?(Rails)
```
Edouard-chin: Rafael was mentioning that it's better to check for Rails::Railtie as Rails itself is sometimes manually defined inside non Rails application

casperisfine: Interesting,[ the doc state to check for `defined?(Rails)`](https://api.rubyonrails.org/classes/Rails/Railtie.html), but I guess we could update it.

Do you mind if I merge as is, and eventually update after with the doc when Rafael get back? I need that change for a cleanup.

Edouard-chin: Yeah sure , let's merge

rafaelfranca: Yeah, let's update the doc.

If a gem is called rails-foo it will define `Rails` even if the rails framework is not required.

casperisfine: https://github.com/rails/rails/pull/37127

carlosantoniodasilva: Interesting, I never hit any issue with `defined?(Rails)` for loading a Railtie before... any chance of `Rails::Railtie` not being required/defined by the time the gem loads, while `Rails` would be? I think not, but just wondering if that couldn't cause loading edge cases.

eugeneius: `require "rails"` [loads `rails/application`](https://github.com/rails/rails/blob/6d68bb5f695414b1801c1c3952ef955a5b0b6c6a/railties/lib/rails.rb#L13), which [loads `rails/engine`](https://github.com/rails/rails/blob/6d68bb5f695414b1801c1c3952ef955a5b0b6c6a/railties/lib/rails/application.rb#L11), which [loads `rails/railtie`](https://github.com/rails/rails/blob/6d68bb5f695414b1801c1c3952ef955a5b0b6c6a/railties/lib/rails/engine.rb#L3). If `Rails` is defined by the `rails` gem, `Rails::Railtie` will be defined too.

casperisfine: What @rafaelfranca suggests here are gems named `rails-somestuf`, with the following namespace:

```ruby
module Rails
  module SomeStuff
  end
end
```

They might or might not `require "rails"`. 

Honestly I've never been bit by this, but I get what Rafael means.

casperisfine: nvm, I misunderstood the question you were answering to.

rafaelfranca: Yeah. The rails-html-sanitizer gem already caused this project in a Sinatra app once. I remember the person even opened an issue on Rails about it.

carlosantoniodasilva: Right yeah, I can totally see how it could happen, just haven't been bitten by it before. Sounds good then, if Rails::Railtie is always there when rails is required. :+1:

PR #393, opened 2019-10-23
Add support for id embedding for has_one caches

gmcgibbon:
Adds support for normalized caching of `has_one` associations via `cache_has_one :name, embed: :id`. This works essentially the same as `embed: :ids` on `cache_has_many` but for one Active Record object and one ID.

### DIFF
--- a/lib/identity_cache/configuration_dsl.rb
+++ b/lib/identity_cache/configuration_dsl.rb
@@ -149,8 +149,11 @@ module IdentityCache
         options = { embed: embed, inverse_name: inverse_name }
         self.cached_has_ones[association] = options
 
-        if options[:embed] == true
+        case embed
+        when true
           build_recursive_association_cache(association, options)
+        when :id
+          build_id_embedded_has_one_cache(association, options)
         else
           raise NotImplementedError
         end
@@ -259,6 +262,42 @@ module IdentityCache
 
           def #{options[:prepopulate_method_name]}(records)
             #{options[:records_variable_name]} = records
+          end
+        CODE
+
+        options[:only_on_foreign_key_change] = true
+        ParentModelExpiration.add_parent_expiry_hook(options)
+      end
+
+      def build_id_embedded_has_one_cache(association, options) #:nodoc:
+        options[:association_reflection]  = reflect_on_association(association)
+        options[:cached_accessor_name]    = "fetch_#{association}"
+        options[:id_name]                 = "#{association}_id"
+        options[:cached_id_name]          = "fetch_#{options[:id_name]}"
+        id_cached_reader_name             = "cached_#{options[:id_name]}"
+        options[:id_variable_name]        = :"@#{id_cached_reader_name}"
+        options[:records_variable_name]   = :"@cached_#{association}"
+        options[:prepopulate_method_name] = "prepopulate_fetched_#{association}"
+
+        self.class_eval(<<-CODE, __FILE__, __LINE__ + 1)
+          attr_reader :#{id_cached_reader_name}
+
+          def #{options[:cached_id_name]}
+            return #{options[:id_variable_name]} if defined?(#{options[:id_variable_name]})
+            #{options[:id_variable_name]} = association(:#{association}).scope.ids.first
+          end
+
+          def #{options[:cached_accessor_name]}
+            association_klass = association(:#{association}).klass
+            if association_klass.should_use_cache? && !association(:#{association}).loaded?
+              #{options[:records_variable_name]} ||= #{options[:association_reflection].class_name}.fetch(#{options[:cached_id_name]}) if #{options[:cached_id_name]}
+            else
+              #{association}
+            end
+          end
+
+          def #{options[:prepopulate_method_name]}(record)
+            #{options[:records_variable_name]} = record
           end
         CODE
 
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -135,6 +135,11 @@ module IdentityCache
             record.instance_variable_set(record.class.cached_has_manys[name][:ids_variable_name], ids)
           end
         end
+        if coder.has_key?(:association_id)
+          coder[:association_id].each do |name, id|
+            record.instance_variable_set(record.class.cached_has_ones[name][:id_variable_name], id)
+          end
+        end
         record.readonly! if IdentityCache.fetch_read_only_records
         record
       end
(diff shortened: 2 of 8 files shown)


lib/identity_cache/query_api.rb
```diff
@@ -468,7 +485,23 @@ def prefetch_one_association(association, records)
             prefetch_embedded_association(records, association, details)
             parent_records = records.map(&details[:cached_accessor_name].to_sym).compact
           else
-            raise ArgumentError.new("Non-embedded has_one associations do not support prefetching yet.")
+            ids_to_parent_record = records.each_with_object({}) do |record, hash|
+              child_ids = Array(record.send(details[:cached_id_name]))
+              child_ids.each do |child_id|
```
dylanahsmith: This makes it seem like `record.send(details[:cached_id_name])` might return an array, which I didn't think was true.  I would think that `if child_id = record.send(details[:cached_id_name]` would make more sense than these two lines.

lib/identity_cache/query_api.rb
```diff
@@ -468,7 +485,23 @@ def prefetch_one_association(association, records)
...
+                hash[child_id] = record
+              end
+            end
+
+            parent_record_to_child_records = Hash.new { |h, k| h[k] = [] }
```
dylanahsmith: Shouldn't the parent just map to a single child record?  If there are multiple then I think active record just uses the first one.  So I think we can just use a `parent_record_to_child_record = {}` hash and use `parent_record_to_child_record[parent_record] ||= child_record` to assign only one record to the parent.

lib/identity_cache/query_api.rb
```diff
@@ -468,7 +485,23 @@ def prefetch_one_association(association, records)
...
+            child_records = details[:association_reflection].klass.fetch_multi(*ids_to_parent_record.keys)
+            child_records.each do |child_record|
+              parent_record = ids_to_parent_record[child_record.id]
+              parent_record_to_child_records[parent_record] << child_record
+            end
+
+            parent_record_to_child_records.each do |parent, children|
+              parent.send(details[:prepopulate_method_name], children)
```
dylanahsmith: This is wrong, it is preloading an array for a cache_has_one association.  This indicates that we are also missing a test for this.

lib/identity_cache/query_api.rb
```diff
@@ -193,6 +198,11 @@ def add_cached_associations_to_coder(record, coder)
             hash[name] = record.instance_variable_get(options[:ids_variable_name]) unless options[:embed] == true
           end
         end
+        if (cached_has_ones = klass.cached_has_ones).any? { |name, options| options[:embed] != true }
```
fbogsany: Is this an optimization assuming that `options[:embed] == :ids` will be uncommon? I understand we're trying to avoid allocations, but the combination of this line and the `unless options[:embed] == true` 2 lines below is jarring. I don't suppose there's a convenient way to filter once?

lib/identity_cache/query_api.rb
```diff
@@ -468,7 +485,21 @@ def prefetch_one_association(association, records)
             prefetch_embedded_association(records, association, details)
```
fbogsany: I'm curious about why this line is inside the `if details[:embed] == true` block for `cached_has_ones` but before the `if details[:embed] == true` for `cached_has_manys`. Which is correct, and if they're both correct, why does it need to be different in each case?

lib/identity_cache/configuration_dsl.rb
```diff
@@ -266,6 +269,42 @@ def #{options[:prepopulate_method_name]}(records)
...
+            return #{options[:ids_variable_name]} if defined?(#{options[:ids_variable_name]})
+            #{options[:ids_variable_name]} = association(:#{association}).scope.ids.first
+          end
+
+          def #{options[:cached_accessor_name]}
+            association_klass = association(:#{association}).klass
+            if association_klass.should_use_cache? && !association(:#{association}).loaded?
+              #{options[:records_variable_name]} ||= #{options[:association_reflection].class_name}.fetch(#{options[:cached_id_name]}) if #{options[:cached_id_name]}
+            else
+              #{association}
```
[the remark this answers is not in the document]
gmcgibbon: `association(:#{association})` returns the association object that we can ask for the AR relation from (and do a `SELECT id ...`) whereas `association` returns the AR object or `nil` from the association (which does a `SELECT * ...`). Has manys return an AR relation via `association` so it isn't needed.

lib/identity_cache/query_api.rb
gmcgibbon: Without this I think we have to bump the IDC cache version because it changes the shape of all coders / serialized payloads by inserting a `association_id` key and a blank hash value.

dylanahsmith: If anything, it would help avoid bloating the cache blob, which might matter for our large shop blob.

lib/identity_cache/query_api.rb
```diff
@@ -193,6 +198,11 @@ def add_cached_associations_to_coder(record, coder)
             hash[name] = record.instance_variable_get(options[:ids_variable_name]) unless options[:embed] == true
           end
         end
+        if (cached_has_ones = klass.cached_has_ones).any? { |name, options| options[:embed] != true }
+          coder[:association_id] = cached_has_ones.each_with_object({}) do |(name, options), hash|
```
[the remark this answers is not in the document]
gmcgibbon: I had originally implemented `id_variable_name` but was persuaded to change it by https://github.com/Shopify/identity_cache/pull/393#discussion_r338302266 to get cache invalidation working quicker. I can revert back though and add to that array if that's better.

lib/identity_cache/query_api.rb
gmcgibbon: Both seem to work, and I have no idea why. I'll make a note to investigate this in another PR.

lib/identity_cache/query_api.rb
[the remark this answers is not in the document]
dylanahsmith: I don't have as strong an opinion on `id_variable_name`, as long as `reload` clears it properly.  However, we are already using the `records_variable_name` name consistently across cache_has_one and cache_has_many recursively embedded associations, so I don't think it should differ for only cache_has_one with id embedding.

lib/identity_cache/query_api.rb
dylanahsmith: I think that might matter in the cases where we are calling `prefetch_associations` on already loaded records (i.e. not through the `includes` `fetch_multi` option).  If that is done on records that wasn't loaded from the cache, then it will refetch the record to use its embedded associations, which would include embedded ids.  So by doing this conditionally, `record.send(details[:cached_id_name])` will load the data from the database.

We should probably be consistent here and make this unconditional as well.

lib/identity_cache/query_api.rb
fbogsany: I *think* the shop blob is one of the first candidates for the use of this mechanism? I agree with avoiding the IDC cache version bump and blob bloat, if possible.

More concretely, though, my comment relates to how the code reads rather than its effect on serialized payload shape. I wonder whether we can filter once, on the line above, and not need to also filter on line 203 below.

gmcgibbon: Changed to a `select` and `present?` check, which only iterates the hash once.

PR #395, opened 2019-10-31
Add :id to cache key hashing for ID embedded hash_ones

gmcgibbon:
Changes string that gets hashed into the IDC cache key for ID embedded has_one associations.

### DIFF
--- a/lib/identity_cache/cache_key_generation.rb
+++ b/lib/identity_cache/cache_key_generation.rb
@@ -17,6 +17,8 @@ module IdentityCache
             schema_string << ",#{name}:(#{denormalized_schema_hash(options[:association_reflection].klass)})"
           when :ids
             schema_string << ",#{name}:ids"
+          when :id
+            schema_string << ",#{name}:id"
           end
         end
       end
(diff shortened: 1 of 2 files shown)


test/cache_key_test.rb
```diff
@@ -0,0 +1,82 @@
+require 'test_helper'
+
+module IdentityCache
+  class CacheKeyTest < IdentityCache::TestCase
+    def test_cache_key
+      assert_equal(
+        "IDC:7:blob:AssociatedRecord:4733125215075107876:1",
```
dylanahsmith: I think it would be clearer to test the string before hashing rather than after hashing.  That way it will be clearer what has changed in the test code.  It looks like you can do that by testing `CacheKeyGeneration.denormalized_schema_hash`.

If you want to test rails_cache_key, then you can do that using `denormalized_schema_hash` on the expected value (e.g. `"IDC:7:blob:AssociatedRecord:#{CacheKeyGeneration.denormalized_schema_hash(klass)}"`) to simply test `rails_cache_key` without testing the different code paths in `denormalized_schema_hash`.

gmcgibbon: I've added `IdentityCache::CacheKeyGeneration.denormalized_schema_string` to expose the unhashed values and tested against that.

PR #397, opened 2019-10-31
Serialize ID embedded has_manys only when present

gmcgibbon:
Currently `coder[:association_ids]` is always set on the serialized coder, even when the value is empty. I think it makes sense to omit it from the serialized payload when we can to save space.


lib/identity_cache/query_api.rb
```diff
@@ -180,7 +180,7 @@ def coder_from_record(record) #:nodoc:
           coder = {
             attributes: record.attributes_before_type_cast.dup,
           }
-          coder[:class] = record.class.name if IdentityCache::CACHE_VERSION == 7
+          coder[:class] = record.class.name
```
gmcgibbon: I don't understand why this was under a conditional before. I'm assuming the constant won't change?

dylanahsmith: This entry is no longer needed.  The intention was to remove it when the cache version got bumped.  I guess I should have added a comment making that clearer.

PR #404, opened 2019-11-09
Move hydration/encoding to separate module

gmcgibbon:
Moves `instrumented_coder_from_record`, `coder_from_record`, `instrumented_record_from_coder `, and `record_from_coder` to separate module.

### DIFF
--- a/lib/identity_cache.rb
+++ b/lib/identity_cache.rb
@@ -3,6 +3,7 @@ require 'active_support/core_ext/module/attribute_accessors'
 require 'ar_transaction_changes'
 
 require "identity_cache/version"
+require "identity_cache/encoder"
 require "identity_cache/cached"
 require "identity_cache/cached/association"
 require "identity_cache/cached/recursive/association"
--- a/lib/identity_cache/encoder.rb
+++ b/lib/identity_cache/encoder.rb
@@ -1,0 +1,88 @@
+module IdentityCache
+  module Encoder
+    DEHYDRATE_EVENT = "dehydration.identity_cache"
+    HYDRATE_EVENT   = "hydration.identity_cache"
+
+    class << self
+      def encode(record)
+        return unless record
+
+        ActiveSupport::Notifications.instrument(DEHYDRATE_EVENT, class: record.class.name) do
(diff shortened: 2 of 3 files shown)


lib/identity_cache/serializer.rb
```diff
@@ -0,0 +1,83 @@
+module IdentityCache
+  module Serializer # :nodoc:
```
dylanahsmith: Prefer using `private_constant` instead of using a comment to indicate that it is private.  That way it is more obvious that privacy is being worked around in code using the library.  That way you could also avoid using `# :nodoc:` on the methods in this module as well.

lib/identity_cache/serializer.rb
```diff
@@ -0,0 +1,83 @@
...
+    DEHYDRATE_EVENT = "dehydration.identity_cache"
+    HYDRATE_EVENT   = "hydration.identity_cache"
```
dylanahsmith: It looks like we are being inconsistent on naming here.  The records aren't actually being serialized to or deserialized from a sequence of bytes, so this isn't really a serializer.  That's why the terms hydration and dehydration were used, even though they aren't as familiar.

lib/identity_cache/serializer.rb
```diff
@@ -0,0 +1,83 @@
...
+          record_from_coder(coder, klass)
+        end
+      end
+
+      private
+
+      def coder_from_record(record, klass)
+        return unless record
+
+        {}.tap do |coder|
```
dylanahsmith: Why not just use

```suggestion
        coder = {}
```

and add `coder` at the end of the method?  That would make the code simpler and reduce the indentation of these already long lines.

gmcgibbon: Stylistic preferences I suppose. Still, I know using tap is sometimes polarizing so I've reverted to using regular assignment in both `coder_from_record` and `record_from_coder`.

lib/identity_cache/serializer.rb
gmcgibbon: Switched to `Encoder`

PR #407, opened 2019-11-14
Prefetch refactor

gmcgibbon:
Moves prefetching to `Cached::Prefetcher` class. This is mostly just moving code around and renaming private methods. We had also discussed moving embedded association prefetching to the same level for both `cached_has_one` and `cached_has_many` (see placement of `prefetch_embedded` calls). I will tackle that next.


lib/identity_cache/cached/prefetcher.rb
```diff
@@ -0,0 +1,200 @@
+# frozen_string_literal: true
+
+module IdentityCache
+  module Cached
+    module Prefetcher
```
rafaelfranca: Is this module public API?

gmcgibbon: Cached is a private constant, so I believe anything under it will also be private.

rafaelfranca: 👍

lib/identity_cache/cached/prefetcher.rb
```diff
@@ -0,0 +1,200 @@
...
+              parent_record = ids_to_parent_record[child_record.id]
+              parent_record_to_child_record[parent_record] ||= child_record
+            end
+
+            parent_record_to_child_record.each do |parent, child|
+              parent.instance_variable_set(cached_has_one.records_variable_name, child)
+            end
+          end
+
+          parent_records
```
dylanahsmith: I think we might have the naming backwards here and the associated records should be referred to as child records rather than parent records.  Looks like that was already the case with the code on master though, so I guess changing it is out of scope of this PR.

If we want a generic term to refer to them across types of associations, then we could refer to them as associated records in the future.

gmcgibbon: Sure, I'll address this in another PR

lib/identity_cache/cached/prefetcher.rb
```diff
@@ -0,0 +1,200 @@
...
+            ActiveRecord::Associations::Preloader.new.preload(records, association)
+            return
+          end
+
+          if (cached_has_many = klass.cached_has_manys[association])
+            fetch_has_many(cached_has_many, records)
+          elsif (cached_belongs_to = klass.cached_belongs_tos[association])
+            fetch_belongs_to(cached_belongs_to, records)
+          elsif (cached_has_one = klass.cached_has_ones[association])
+            fetch_has_one(cached_has_one, records)
```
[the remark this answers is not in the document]
gmcgibbon: Great idea, that would simplify this module by a lot! I'll tackle that and your other point next.

PR #408, opened 2019-11-14
Move fetching to cached association methods

gmcgibbon:
Moves fetching to `Cached::EmbeddedFetching` module and rename `Cached::Reference::BelongsTo` to `Cached::BelongsTo`


lib/identity_cache/cached/association.rb
```diff
@@ -39,6 +39,42 @@ def inverse_name
         end
       end
 
+      def fetch(records)
+        return if fetched?(records)
```
gmcgibbon: `fetched?` runs the same checks as `prefetch_embedded` before calling `fetch_embedded` which is now below 👇

lib/identity_cache/cached/association.rb
```diff
@@ -39,6 +39,42 @@ def inverse_name
...
+              record.instance_variable_set(cached_association.records_variable_name, records)
+            end
+          end
+        end
+      end
+
+      def fetched?(records)
+        # Assume all records are the same, so just check the first one.
+        record = records.first
+        record.association(name).loaded? || record.instance_variable_defined?(records_variable_name)
```
gmcgibbon: `record.instance_variable_defined?(records_variable_name)` wasn't in the base check before, but we can assume all associations have this, and that it is loaded when it exists.

lib/identity_cache/cached/reference/has_one.rb
```diff
@@ -41,6 +41,34 @@ def clear(record)
...
+          parent_record_to_child_record.each do |parent, child|
+            parent.instance_variable_set(records_variable_name, child)
+          end
+
+          child_records
+        end
+
+        def fetched?(records)
+          record = records.first
+          super || record.instance_variable_defined?(id_variable_name)
```
gmcgibbon: `record.instance_variable_defined?(id_variable_name)` was missing before, so I don't think load checks on has_ones were working.

lib/identity_cache/cached/association.rb
```diff
@@ -39,6 +39,42 @@ def inverse_name
         end
       end
 
+      def fetch(records)
```
gmcgibbon: `fetch` now runs before every fetch type whereas before it only ran for `has_many`s and recursive `has_one`s. I don't see a problem with this.

lib/identity_cache/cached/association.rb
```diff
@@ -39,6 +39,42 @@ def inverse_name
...
+              record.instance_variable_set(cached_association.ids_variable_name, ids)
+            when Cached::Reference::HasOne
+              id = cached_record.public_send(cached_association.cached_id_name)
+              record.instance_variable_set(cached_association.id_variable_name, id)
+            else
+              records = cached_record.public_send(cached_association.cached_accessor_name)
+              record.instance_variable_set(cached_association.records_variable_name, records)
+            end
+          end
+        end
```
dylanahsmith: The advantage of moving the fetching to the cached association class is that we could avoid a lot of type checking that is still being done here.  Instead, the type specific logic can be moved into the derived cached association class that actually needs it.

For instance, none of this code makes sense for a cached belongs to association.

lib/identity_cache/cached/prefetcher.rb
```diff
@@ -44,155 +44,12 @@ def fetch_association(klass, association, records)
...
-              parent_records.each do |parent_record|
-                child_record = ids_to_child_record[parent_record.id]
-                child_record.instance_variable_set(cached_belongs_to.records_variable_name, parent_record)
-              end
-            end
-          end
+          cached_association = klass.cached_has_manys[association] ||
+                               klass.cached_has_ones[association] ||
+                               klass.cached_belongs_tos[association] ||
+                               raise(ArgumentError, "Unknown cached association #{association} listed for prefetching")
```
dylanahsmith: These lines could be replaced with `klass.cached_association(association)`.

lib/identity_cache/cached/prefetcher.rb
```diff
@@ -44,10 +44,9 @@ def fetch_association(klass, association, records)
             return
           end
 
-          cached_association = klass.cached_has_manys[association] ||
-                               klass.cached_has_ones[association] ||
-                               klass.cached_belongs_tos[association] ||
-                               raise(ArgumentError, "Unknown cached association #{association} listed for prefetching")
+          cached_association = klass.send(:cached_association, association) || raise(
```
dylanahsmith: It will actually use Hash#fetch for the last lookup in `cached_association` to raise an error instead of returning `nil`, so this `|| raise` is dead code right now.

lib/identity_cache/cached/recursive/association.rb
```diff
@@ -47,6 +52,32 @@ def embedded_by_reference?
...
         end
+
+        private
+
+        def fetched?(records)
+          record = records.first
+          super || record.instance_variable_defined?(dehydrated_variable_name)
+        end
+
+        def fetch_embedded(records)
```
dylanahsmith: This isn't just used for recursively embedded associations.  It is also used for id embedded associations to get the associated ids.

lib/identity_cache/cached/prefetcher.rb
```diff
@@ -44,155 +44,8 @@ def fetch_association(klass, association, records)
...
-              when Cached::Reference::HasOne
-                cached_association = cached_record.public_send(association.cached_id_name)
-                record.instance_variable_set(association.id_variable_name, cached_association)
-              else
-                cached_association = cached_record.public_send(association.cached_accessor_name)
-                record.instance_variable_set(association.records_variable_name, cached_association)
-              end
-            end
-          end
+          cached_association = klass.send(:cached_association, association)
```
rafaelfranca: Why are we using `send` for a method that we own? If it is to "tell" people this method is private API, documentation would be better. We can say that people will ignore the documentation, but if they want to use this method even if the documentation say it is private they will likely also use `send`.

gmcgibbon: I'd be in favour of making this a nodoc public method, but I was planning on doing this later. I'm trying my best to improve IDC's codebase little by little 🙃

gmcgibbon: https://github.com/Shopify/identity_cache/pull/410

lib/identity_cache/cached/recursive/association.rb
```diff
@@ -24,19 +24,40 @@ def #{cached_accessor_name}
...
         def clear(record)
           if record.instance_variable_defined?(records_variable_name)
             record.remove_instance_variable(records_variable_name)
           end
         end
 
+        def fetch(records)
+          fetch_embedded(records)
+
+          records.flat_map(&cached_accessor_name.to_sym).tap(&:compact!)
```
dylanahsmith: We should move the `.to_sym` to `Cached::Association#initialize` so that we aren't always converting it to a symbol at runtime.  It is only used as a string as part of the class eval strings.  Doesn't need to happen in this PR though.

gmcgibbon: 👌 , I'll make that change and update the `cached_association` call

PR #409, opened 2019-11-15
Restore IdentityCache::QueryAPI#prefetch_associations

gmcgibbon:
Restores `prefetch_associations`, its public API.


rafaelfranca:
Should we add tests?

gmcgibbon:
Yes, I added comments so its clearer that it is public, but I'll add a test too to prevent regression.

PR #410, opened 2019-11-15
Make IdentityCache::QueryAPI#cached_association public

gmcgibbon:
Makes `cached_association` a `:nodoc:` public method.

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -100,6 +100,11 @@ module IdentityCache
         IdentityCache.cache.delete(rails_cache_key(id))
       end
 
+      # @api private
+      def cached_association(name) # :nodoc:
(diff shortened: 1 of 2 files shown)


lib/identity_cache/query_api.rb
```diff
@@ -100,6 +100,10 @@ def expire_primary_key_cache_index(id)
         IdentityCache.cache.delete(rails_cache_key(id))
       end
 
+      def cached_association(name) # :nodoc:
```
rafaelfranca: We can also add a comment like `# @api private` on top just to make it clear.

dylanahsmith: Yeah, I always found `:nodoc:` to be kind of implicit about its meaning

gmcgibbon: TIL 👍

PR #412, opened 2019-11-19
Remove require_if_necessary, since it is no longer necessary

dylanahsmith:
I think this was added to support loading constants that hadn't been autoloaded yet, back when we marshal dumped the record objects themselves and before [active support patched Marshal.load to do this](https://github.com/rails/rails/blob/v6.0.1/activesupport/lib/active_support/core_ext/marshal.rb).

This is no longer necessary, because we cache the attributes before they are type cast to avoid making them dependent on active record's object format or any constant names.

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -26,23 +26,21 @@ module IdentityCache
         id = type_for_attribute(primary_key).cast(id)
         return unless id
         record = if should_use_cache?
-          require_if_necessary do
-            object = nil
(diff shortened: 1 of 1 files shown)


camilo:
This is really really really old code so yes +1 for 🔥 .  The roll out of a version with this patch in say a large older code base needs to be probably done very careful in case that proverbial codebase would need this code path for some other reason.

dylanahsmith:
@camilo We haven't been caching constant names for a long time, so I don't think it should matter even for a large older code base.  This isn't even dependant on the active support Marshal.load patch.

PR #413, opened 2019-11-21
Prefetch with batch operation objects.

gmcgibbon:
Uses batch operation objects to prefetch associations. Later, we should be able to load each batch with an abstract `fetch_multi` interface to reduce round-trips to memcached.


rafaelfranca:
Looks like this code is now instantiating way more objects than the previous code. Is it really worth? For example does `Operation` need to be a object?

gmcgibbon:
> Is it really worth?

Currently, not at all, but I think we _do_ need something to parse includes syntax into something that can be easily batched. This will inevitably require more object allocations. Since we're not batching yet, it is probably difficult to see the value. 

> For example does `Operation` need to be a object?

There is very likely lots room for improvement in my implementation (and perhaps my mental model of how batching should work in IDC). I'll see what I can do.

rafaelfranca:
Let's move on with this since it can allow you to make other changes, but we should see later if we can remove some allocations and measure.

PR #414, opened 2019-11-25
Avoid exposing primary index methods on models including WithoutPrimaryIndex

dylanahsmith:
## Problem

IdentityCache::WithoutPrimaryIndex was including the `IdentityCache` module then disabling some of its functionality by setting a `primary_cache_index_enabled` to false to cause methods to raise.  This is the opposite of how inheritance is normally done, where a module or derived class should extend the ancestors while preserving its behaviour.  We also can't use an `is_a?` check to see if functionality exists since `primary_cache_index_enabled` could be changed to not reflect the modules that are included.

## Solution

I moved methods that depend on the primary cache index to a `IdentityCache::WithPrimaryIndex` module, which includes `IdentityCache::WithoutPrimaryIndex` instead of the reverse.  The `IdentityCache` module includes `WithPrimaryIndex`.

### DIFF
--- a/lib/identity_cache/with_primary_index.rb
+++ b/lib/identity_cache/with_primary_index.rb
@@ -1,0 +1,203 @@
+# frozen_string_literal: true
+module IdentityCache
+  module WithPrimaryIndex
+    extend ActiveSupport::Concern
+
+    include WithoutPrimaryIndex
+
+    def expire_cache
+      expire_primary_index
+      super
+    end
+
+    # @api private
+    def expire_primary_index # :nodoc:
+      self.class.expire_primary_key_cache_index(id)
+    end
+
+    # @api private
+    def primary_cache_index_key # :nodoc:
+      self.class.rails_cache_key(id)
+    end
+
+    module ClassMethods
+      def primary_cache_index_enabled
+        true
+      end
+
+      # Declares a new index in the cache for the class where IdentityCache was
+      # included.
+      #
+      # IdentityCache will add a fetch_by_field1_and_field2_and_...field for every
+      # index.
+      #
+      # == Example:
+      #
+      #  class Product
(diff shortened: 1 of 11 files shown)


lib/identity_cache/with_primary_index.rb
```diff
@@ -0,0 +1,203 @@
+# frozen_string_literal: true
+module IdentityCache
+  module WithPrimaryIndex
+    extend ActiveSupport::Concern
+
+    include WithoutPrimaryIndex
```
gmcgibbon: I think this approach makes sense from a reusability standpoint, but I don't like that `IdentityCache` modules include both `WithPrimaryIndex` and `WithoutPrimaryIndex`. That said, I do agree with the creation of this module. 

We should work toward having `WithPrimaryIndex` and `WithoutPrimaryIndex` being our base includes and having `include IdentityCache` be an alias for `include IdentityCache::WithPrimaryIndex` for legacy reasons. In the meantime, this is still better than the previous code, so I'll approve. 👍

dylanahsmith: I agree that the `WithoutPrimaryIndex` name is kind of weird, once `WithPrimaryIndex` also gets included.  Maybe `WithoutPrimaryIndex` should be renamed to `Base` in the future so it isn't implying something with the name that could become false by including another module.

I would like to avoid `include IdentityCache` in the future, since that also includes its constants into the module.  I think it would be better to use `include IdentityCache::Base` and then call `cache_primary_index` to opt-in to using the primary index, which could internally include a module to make primary index methods available.  Calling a method to opt-in to the primary index would also allow us to pass in options, like explicitly specifying the embedded associations in one place, so that it is clear how much we are embedding.

PR #415, opened 2019-11-25
Move primary cache index implementation to a Cached::PrimaryIndex class

dylanahsmith:
I would like to have similar objects for different cached associations, including the primary index.  Creating a separate object for the primary index will make it easier to provide a common interface for cache fetching to allow easier re-use between the primary index and attribute indexes without also polluting the model with private fetch methods.

This PR just does the implementation extraction and doesn't do the same for attribute indexes or try to share functionality with attribute indexes to keep the PR simpler.


lib/identity_cache/with_primary_index.rb
```diff
@@ -17,10 +17,15 @@ def expire_primary_index # :nodoc:
 
     # @api private
     def primary_cache_index_key # :nodoc:
-      self.class.rails_cache_key(id)
+      self.class.cached_primary_index.cache_key(id)
     end
 
     module ClassMethods
+      # @private
```
gmcgibbon:
```suggestion
      # @api private
```

lib/identity_cache/cached/primary_index.rb
```diff
@@ -0,0 +1,114 @@
+# frozen_string_literal: true
+
+module IdentityCache
+  module Cached
+    class PrimaryIndex
```
gmcgibbon: You should update your commit message to mention this class name and not `PrimaryKey`.

dylanahsmith:
@gmcgibbon addressed your feedback

PR #416, opened 2019-11-25
Extract cache attribute implementation into a separate class

dylanahsmith:
The reason for doing this is the same as https://github.com/Shopify/identity_cache/pull/415 and follows the pattern we have done for extracting cached associations to their own class.


lib/identity_cache/cached/attribute.rb
```diff
@@ -0,0 +1,158 @@
+# frozen_string_literal: true
+
+module IdentityCache
+  module Cached
+    class Attribute
+      attr_reader :model, :attribute, :alias_name, :key_fields, :unique
+
+      def initialize(model, attribute, alias_name, key_fields, unique)
+        @model = model
+        @attribute = attribute
```
gmcgibbon: Does this need to be symbolized too?

dylanahsmith: It wasn't before, but it would probably be better to be clear about its type.

PR #419, opened 2019-12-20
Refactor cached attribute to handle composite keys separately

dylanahsmith:
##  Problem

I would like a common cache fetcher interface that can be implemented for the primary index and cache attribute fetches.  However, loading by a single key or a multiple key works differently for cached attributes, since it doesn't support a fetch multi by a composite key. 

## Solution

Split Cached::Attribute into Cached::AttributeByOne and Cached::AttributeByMulti, sharing a common base class.  Only Cached::AttributeByOne of these supports fetch_multi.


lib/identity_cache/configuration_dsl.rb
```diff
@@ -131,7 +131,8 @@ def cache_attribute_by_alias(attribute_or_proc, alias_name:, by:, unique:)
         ensure_base_model
         fields = Array(by)
 
-        cached_attribute = Cached::Attribute.new(self, attribute_or_proc, alias_name, fields, unique)
+        klass = fields.length == 1 ? Cached::AttributeByOne : Cached::AttributeByMulti
```
gmcgibbon:
```suggestion
        klass = fields.one? ? Cached::AttributeByOne : Cached::AttributeByMulti
```

lib/identity_cache/cached/attribute.rb
```diff
@@ -90,42 +46,38 @@ def expire(record)
...
-          rows.each do |index_value, attribute_value|
-            result[index_value] = attribute_value
-          end
-        else
-          rows.each do |index_value, attribute_value|
-            result[index_value] << attribute_value
-          end
-        end
-        result
+      private
```
gmcgibbon: If we expect these methods to be overridden by the child and they aren't public, should we make them `protected` instead.

lib/identity_cache/cached/attribute_by_one.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+
+module IdentityCache
+  module Cached
+    class AttributeByOne < Attribute
```
gmcgibbon: Should we rename these to `Attribute::One` and `Attribute::Many`, or maybe `SingleAttribute` and `BatchAttribute`? `*ByOne` and `*ByMulti` seems very fixated on the implementation to me. WDYT?

lib/identity_cache/cached/attribute.rb
dylanahsmith: In ruby, `protected` is to allow them to be accessed from another instance.  If their purpose isn't clear enough, maybe I could make that clearer with documentation.  Looks like YARD has an [`@abstract` tag](https://rubydoc.info/gems/yard/file/docs/Tags.md#abstract) that could be used for that purpose.

lib/identity_cache/cached/attribute_by_one.rb
dylanahsmith: The `By` word is important, because this is referring to the key.  It actually isn't fixated on the caching implementation, it is meant to correspond with the usage. E.g. `cache_attribute :quantity, by: :name` uses Cached::AttributeByOne and `cache_attribute :quantity, by: [:name, :vendor]` uses Cached::AttributeByMulti.  In both cases they are only caching a single attribute, although it wouldn't be hard to support multiple attributes in the future.

lib/identity_cache/cached/attribute.rb
gmcgibbon: Yeah, that should work too. I don't typically see abstract methods that are private though I'm sure they exist.

garethson:
@dylanahsmith re: 
> Loading by a single key or a multiple key works differently for cached attributes, since it doesn't support a fetch multi by a composite key.

I'm looking for this functionality, specifically:
```ruby
cache_attribute :status, by: [:id, :token], unique: false
```
and then `Item.fetch_multi_status_by_id_and_token`

Are you open to me adding a to PR to introduce this feature?  Looking at when fetch_multi support for cache_index for a single key [was added](https://github.com/Shopify/identity_cache/pull/368), would you caution me from going down this path, or is it something that on the surface seems reasonable for me to investigate? I currently have little knowledge of this gem's internals, so just looking for feedback on the concept before I dive in, which I'm more than happy to.

dylanahsmith:
@garethson In order to support fetch multi cache attribute by multiple columns, we basically would want a tuple comparison query.  It looks like Active Record didn't provide support for it and attempts to add support for that upstream seem to have been closed (https://github.com/rails/rails/issues/40273 and https://github.com/rails/rails/pull/36003) so we can't just do something like this to query the database

```
statuses = Item.where([:some_id, :token] => [[123, "abc"], [234, "bcd"]]).pluck(:status)
```

It looks like we would instead need to build the SQL query string.  E.g. for the above example, we would want a SQL query like

```sql
SELECT status
FROM items
WHERE (some_id, token) IN ((123, 'abc'), (234, 'bcd'))
```

We would want to quote the values, where the simplest approach might be to build the query with a `?` placeholder for each tuple and pass in the splat of the tuples array

```
Item.where("(some_id, token) IN ((?),(?))", *tuples).pluck(:status)
```

I would expect the IDC API to be given that array of tuples as an argument, so the above query would come from the following IDC call

```
Item.fetch_multi_status_by_some_id_and_token([[123, "abc"], [234, "bcd"]])
```

PR #420, opened 2019-12-20
Refactor to separate generic cache key loading logic

dylanahsmith:
## Problem

We want to be able to fetch cache keys across cache fetchers to reduce round trips for id embedded associations.  This means we need a common IdentityCache.fetch_multi, but then different cache keys will need to be resolved by different cache fetchers.

## Solution

Document a common interface for cache fetchers so that they can be used to batch load they cache key fetching together while also being able to resolve misses.  I've moved the generic cache key loading logic into a CacheKeyLoader singleton object, which can be extended with a `batch_load` method once we have some code to use it.

I've opened up the draft PR https://github.com/Shopify/identity_cache/pull/421 showing an implementation of CacheKeyLoader.batch_load.


lib/identity_cache/cache_key_loader.rb
```diff
@@ -0,0 +1,99 @@
+# frozen_string_literal: true
+
+module IdentityCache
+  # A generic cache key loader that supports different types of
+  # cache fetchers, each of which can use their own cache key
+  # format and have their own cache miss resolvers.
+  #
+  # Here is the interface of a cache fetcher in the
+  # [ruby-signature](https://github.com/ruby/ruby-signature)'s
+  # format.
```
gmcgibbon: TIL 👍

lib/identity_cache/cache_key_loader.rb
```diff
@@ -0,0 +1,99 @@
...
+  # interface _CacheFetcher[DbKey, DbValue, CacheableValue]
+  #   def cache_key: (DbKey) -> String
+  #   def cache_encode: (DbValue) -> CacheableValue
+  #   def cache_decode: (CacheableValue) -> DbValue
+  #   def load_one_from_db: (DbKey) -> DbValue
+  #   def load_multi_from_db: (Array[DbKey]) -> Hash[DbKey, DbValue]
+  # end
+  # ```
+  module CacheKeyLoader
+    # Load a single keys for a cache fetcher
```
gmcgibbon:
```suggestion
    # Load a single key for a cache fetcher.
```

lib/identity_cache/cache_key_loader.rb
```diff
@@ -0,0 +1,99 @@
...
+
+      cache_value = IdentityCache.fetch(cache_key) do
+        db_value = cache_fetcher.load_one_from_db(db_key)
+        cache_fetcher.cache_encode(db_value)
+      end
+
+      db_value || cache_fetcher.cache_decode(cache_value)
+    end
+
+    # Load multiple keys for a cache fetcher
```
gmcgibbon:
```suggestion
    # Load multiple keys for a cache fetcher.
    # 
```

lib/identity_cache/cache_key_loader.rb
```diff
@@ -0,0 +1,99 @@
...
+      load_result ||= {}
+      cache_result.each do |cache_key, cache_value|
+        db_key = cache_key_to_db_key_hash.fetch(cache_key)
+        load_result[db_key] ||= cache_fetcher.cache_decode(cache_value)
+      end
+
+      load_result
+    end
+
+    class << self
```
gmcgibbon: Should we put ☝️ in this `class << self` block or is there a reason why we need to use `self.method` syntax for them?

dylanahsmith: yeah, we can put the public singleton methods in the `class << self` block and it won't be any different.

Issue #422, opened 2020-01-06
ActiveRecord::RecordNotFound is raised, but record is exist in database


saiqulhaq:
as the title, why raise ActiveRecord::RecordNotFound if my record in database is exist?

my stack:
Rails 5.0, MySQL 5.7 in AWS

saiqulhaq:
caused by this line https://github.com/Shopify/identity_cache/blob/8f68b3df84e03a4e966a82795edb530f03e79f42/lib/identity_cache/query_api.rb#L45

rafaelfranca:
@dylanahsmith @gmcgibbon maybe we should raise an specific exception that inherits from RecordNotFound just to make clear?

gmcgibbon:
👍

saiqulhaq:
this issue is happening again, I am pointing github master in my Gemfile (identity_cache-0cbe7ec09516)

it says that `IdentityCache::RecordNotFound will be raised if id is not in the cache or the db.`

https://github.com/Shopify/identity_cache/blob/0cbe7ec09516569f79fdee7a3ac25c9e16084229/lib/identity_cache/with_primary_index.rb#L113

but the reality is like this
![image](https://user-images.githubusercontent.com/1275215/72496006-dc275a00-385a-11ea-8b31-4ea31c5341ff.png)

gmcgibbon:
Oh, right. I need to update the docs to demonstrate the new behaviour. `IdentityCache::RecordNotFound` is raised for no cache entry, and `ActiveRecord::RecordNotFound` is raised for no db entry. Thanks for pointing that out!

dylanahsmith:
It sounds like `nil` might have been cached for the record and then not invalidated by the creation of that record.

It is hard to say why the cache wasn't invalidated, but make sure to pay attention to the configuration in the installation section of the README to avoid using a cache backend that doesn't have CAS support and to avoid the `auto_eject_hosts: true` memcached client configuration.  Otherwise, logging might be the best way to try to track down why the cache invalidation might have been lost.

[...]

PR #425, opened 2020-01-20
Fix IdentityCache::WithPrimaryIndex#fetch docs

gmcgibbon:
Clarify exception throwing behaviour in docs of #fetch.

Amending doc changes of the fix for https://github.com/Shopify/identity_cache/issues/422.


saiqulhaq:
thanks

PR #426, opened 2020-01-22
Start batch prefetching non-embedded cached associations

dylanahsmith:
@gmcgibbon Here is the follow-up work from https://github.com/Shopify/identity_cache/pull/421 that starts doing lazy batch loading of non-embedded associations.


dylanahsmith:
Oops, this was supposed to be a draft PR

PR #427, opened 2020-01-22
Start batch prefetching non-embedded cached associations

dylanahsmith:
@gmcgibbon Here is the follow-up work from https://github.com/Shopify/identity_cache/pull/421 that starts doing lazy batch loading of non-embedded associations.


lib/identity_cache/cached/belongs_to.rb
```diff
@@ -60,14 +62,16 @@ def fetch(records)
...
             end
           end
-          parent_records = reflection.klass.fetch_multi(ids_to_child_record.keys)
-          parent_records.each do |parent_record|
-            child_record = ids_to_child_record[parent_record.id]
-            child_record.instance_variable_set(records_variable_name, parent_record)
+
+          load_strategy.load_multi(reflection.klass.cached_primary_index, ids_to_child_record.keys) do |parent_records_by_id|
+            parent_records_by_id.each do |id, parent_record|
+              next unless child_record = ids_to_child_record[id]
```
dylanahsmith: When would `child_record` be `nil`?  Isn't it the owner of the association?  I don't see why we would have loaded a record without an owner.

I think we should be able to do

```suggestion
             child_record = ids_to_child_record.fetch(id)
```

We also have `next unless` in the polymorphic case above, which also seems like it should be unnecessary.

It looks like you also repeated this in reference/has_many.rb and reference/has_one.rb as well.

lib/identity_cache/cached/reference/has_many.rb
```diff
@@ -49,28 +49,33 @@ def clear(record)
...
 
-          parent_record_to_child_records.each do |parent, children|
-            parent.instance_variable_set(records_variable_name, children)
-          end
+              parent_record_to_child_records.each do |parent, children|
+                parent.instance_variable_set(records_variable_name, children)
+              end
 
-          child_records
+              yield child_records_by_id.values
```
dylanahsmith: Doesn't this need to omit any `nil` values?  It looks like this was done since my last commit.

It looks like you repeated this in reference/has_one.rb as well.

gmcgibbon: Right, I can probably get rid of some `nil` compactions / skipping if I omit them here. Also less likely to cache `nil` values this way I assume.

Gemfile
```diff
@@ -4,3 +4,4 @@ gemspec
 
 gem 'mysql2', '~> 0.5.2'
 gem 'rubocop'
+gem 'byebug'
```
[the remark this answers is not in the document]
gmcgibbon: I was tired of adding it to the Gemfile every time I needed to debug something. Can revert, but this seems reasonable as well.

lib/identity_cache/cached/reference/has_many.rb
dylanahsmith: > Also less likely to cache `nil` values this way I assume.

What gets cached is handled internally to the cache fetchers, so there shouldn't be a worry about caching the wrong thing outside of it.  Besides, it actually does cache nil for records that it doesn't find in the database.

Gemfile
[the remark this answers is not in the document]
dylanahsmith: No, I agree we should add it.  I was just suggesting making it clear that it is only for MRI.

lib/identity_cache/cached/belongs_to.rb
```diff
@@ -43,15 +47,13 @@ def fetch(records)
             types_to_parent_ids[parent_type][parent_id] = child_record
           end
 
-          parent_records = []
-
-          types_to_parent_ids.each do |type, ids_to_child_record|
-            type_parent_records = type.fetch_multi(ids_to_child_record.keys)
-            type_parent_records.each do |parent_record|
-              child_record = ids_to_child_record[parent_record.id]
+          load_strategy.load_batch(types_to_parent_ids.transform_keys(&:cached_primary_index)) do |parent_records_by_id|
```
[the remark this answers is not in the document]
gmcgibbon: Seems like we're missing a test for that then.

lib/identity_cache/cached/belongs_to.rb
gmcgibbon: There's a nil value in my prefetch test when prefetching multiple items at different levels. 🤔

dylanahsmith:
> It looks like we could also use shared tests for the load strategy interface to make sure `Eager` and `Lazy` are consistent.

I pushed a commit that did this.

---

:+1: this PR looks good to me now, although being the original PR author, github won't let me give it a review approval

PR #430, opened 2020-01-31
Layout style fixes

gmcgibbon:
Fixes layout style violations. Also fixes rubocop workflow on master.


.github/workflows/ci.yml
```diff
@@ -13,8 +13,14 @@ jobs:
       uses: actions/setup-ruby@v1
       with:
         ruby-version: 2.6.x
-    - name: Ruby style checks
+    - name: Install required packages
+      run: |
+        sudo apt-get update
+        sudo apt-get -y install libmysqlclient-dev libpq-dev libsqlite3-dev libsasl2-dev
```
dylanahsmith: I don't think we test with sqlite3 in CI right now, although I don't see why it wouldn't work.

What is libsasl2-dev a dependency of?

gmcgibbon: Yep, I can remove sqlite. libsasl2 is for memcached.

.github/workflows/ci.yml
```diff
@@ -13,8 +13,14 @@ jobs:
       uses: actions/setup-ruby@v1
       with:
         ruby-version: 2.6.x
```
dylanahsmith: Does this mean that it is only tested with ruby 2.6?  If we don't have any ruby version conditional code, then we should test with the minimum supported (2.4) and latest released (2.7).

dylanahsmith: Oh, this is only being used for rubocop at the moment.  Although, that makes me wonder why we are installing database clients (libmysqlclient-dev libpq-dev) if we don't have a step/job to run the test suite.

gmcgibbon: We are still using travis for test runs because orchestrating a MySQL DB for testing with github actions isn't super straightforward. I thought I'd defer working on a script to do that, and run a matrix of tests/versions for later. The client packages are just so we can `gem install`.

PR #433, opened 2020-02-05
Fix remaining style violations

gmcgibbon:
Fixes all remaining style violations in IDC. ✨


identity_cache.gemspec
```diff
@@ -22,7 +22,9 @@ Gem::Specification.new do |gem|
                       "cannot be found in Memcached."
   gem.homepage      = "https://github.com/Shopify/identity_cache"
 
-  gem.files         = `git ls-files`.split($\)
+  gem.files         = Dir.chdir(File.expand_path("..", __FILE__)) do
+    %x(git ls-files -z).split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
```
dylanahsmith: There aren't top-level `spec` or `features` directories, so why bother adding those?

```suggestion
    %x(git ls-files -z).split("\x0").reject { |f| f.match(%r{^test/}) }
```

identity_cache.gemspec
```diff
@@ -22,7 +22,9 @@ Gem::Specification.new do |gem|
                       "cannot be found in Memcached."
   gem.homepage      = "https://github.com/Shopify/identity_cache"
 
-  gem.files         = `git ls-files`.split($\)
+  gem.files         = Dir.chdir(File.expand_path("..", __FILE__)) do
```
dylanahsmith:
```suggestion
  gem.files         = Dir.chdir(__dir__) do
```

identity_cache.gemspec
gmcgibbon: Same pattern as the one used in `gem.test_files`. I've updated both to be less generic. 👍

PR #435, opened 2020-02-07
Version 1.0.0

gmcgibbon:
#### Release Identity Cache 1.0.0

With batch fetching shipped, I think this is a good time to release a new version. The next release seems to suggest we want to skip all the way to `1.0`. That might be a good idea, since IDC has gone through a rather large amount of changes recently. Please have a look at the diff to see if I missed anything important: https://github.com/Shopify/identity_cache/compare/8f68b3df84e03a4e966a82795edb530f03e79f42...master

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,15 +5,21 @@
 - Remove inverse_name option. Specify inverse_of on the Active Record association instead. (#439)
 - Bump the minimum Active Record version to 5.2 (#438)
 - Remove the default embed option value from cache_has_one (#437)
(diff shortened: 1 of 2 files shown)


dylanahsmith:
This might be a good time to remove the default embed parameter value for cache_has_one so we can discourage recursive embedding by default.  Then we can add it back with a default consistent with cache_has_many

dylanahsmith:
Other possible breaking changes to consider would be:
* getting rid of `inverse_name` in favour of specifying `inverse_of` on the active record association.
* bump the minimum supported active record version to 5.2, since older versions aren't supported upstream

Issue #428, opened 2020-01-27
NoMethodError: undefined method `ids_variable_name' for nil:NilClass


saiqulhaq:
https://github.com/Shopify/identity_cache/blob/3f626e01d3f22ce454a42f29e241719e850512d6/lib/identity_cache/encoder.rb#L79

dylanahsmith:
Could you please provide steps to reproduce the exception?

saiqulhaq:
this is in my Model

parent
![image](https://user-images.githubusercontent.com/1275215/73326897-640f5a00-4286-11ea-83be-cbc40328ead9.png)

child
![image](https://user-images.githubusercontent.com/1275215/73326886-58239800-4286-11ea-860b-0142321496a3.png)

the error is on line number 4 `variant = object.fetch_variant`
![image](https://user-images.githubusercontent.com/1275215/73326920-76899380-4286-11ea-84aa-ab4aee49162d.png)

saiqulhaq:
I don't know how to reproduce, it's popping up randomly once or twice in a day usually

but here is the backtrace

```
NoMethodError: undefined method `ids_variable_name' for nil:NilClass
1 File "/var/www/app/shared/bundle/ruby/2.5.0/bundler/gems/identity_cache-37dc8b8188c4/lib/identity_cache/encoder.rb" line 79 in block in record_from_coder
2 File "/var/www/app/shared/bundle/ruby/2.5.0/bundler/gems/identity_cache-37dc8b8188c4/lib/identity_cache/encoder.rb" line 78 in each
3 File "/var/www/app/shared/bundle/ruby/2.5.0/bundler/gems/identity_cache-37dc8b8188c4/lib/identity_cache/encoder.rb" line 78 in record_from_coder
4 File "/var/www/app/shared/bundle/ruby/2.5.0/bundler/gems/identity_cache-37dc8b8188c4/lib/identity_cache/encoder.rb" line 20 in block in decode
5 File "/var/www/app/shared/bundle/ruby/2.5.0/gems/activesupport-6.0.2.1/lib/active_support/notifications.rb" line 182 in instrument
6 File "/var/www/app/shared/bundle/ruby/2.5.0/bundler/gems/identity_cache-37dc8b8188c4/lib/identity_cache/encoder.rb" line 19 in decode
[...]

PR #436, opened 2020-02-07
Help make error reported in issue #428 clearer

dylanahsmith:
cc @saiqulhaq

The exception reported in issue #428 was ```NoMethodError: undefined method `ids_variable_name' for nil:NilClass``` on line [lib/identity_cache/encoder.rb:79](https://github.com/Shopify/identity_cache/blob/c4799ac93b489312d9685dd674750c93a4729b02/lib/identity_cache/encoder.rb#L79).  That exception indicates that `Hash#[]` returned `nil` instead of the cached association.  Using Hash#fetch instead provides a better error message by also providing the association name, which could help in debugging this possible exception.

I'm not actually sure how the exception could be reproduced, but I did notice that the code was trying to get the cached association from the instantiated record class rather than from the cached model that `fetch_by_id` was called on.  These classes aren't the same when single table inheritance is used, so it would be better to avoid this indirection to any possible effects of this single table inheritance edge case.  However, `cached_has_manys` is a `class_attribute` so it should have been inherited by the child class, so I don't actually have a reason to believe that this caused the exception.

### DIFF
--- a/lib/identity_cache/encoder.rb
+++ b/lib/identity_cache/encoder.rb
@@ -76,12 +76,12 @@ module IdentityCache
         end
         if coder.key?(:association_ids)
           coder[:association_ids].each do |name, ids|
-            record.instance_variable_set(record.class.cached_has_manys[name].ids_variable_name, ids)
+            record.instance_variable_set(klass.cached_has_manys.fetch(name).ids_variable_name, ids)
           end
         end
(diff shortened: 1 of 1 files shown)


saiqulhaq:
yes correct, it's caused by instantiated record
thanks for the PR

PR #437, opened 2020-02-11
Remove the default embed option value from cache_has_one

dylanahsmith:
## Problem

`cache_has_many` and `cache_has_one` have different default values for the `embed` option.  `cache_has_many` defaults to `embed: :ids` and `cache_has_one` defaults to `embed: true`.

We have also run into performance problems from relying too much on embedding, since it can result in a lot of over-fetching from the cache which can cause performance problems for the cache.  It also makes it very easy for other code to rely on the association to be prefetched, which is coupling in the code that can be harder to remove later.

## Solution

Just changing the default to be consistent would significantly change the application and add performance problems that would be annoying to debug.  Just removing the default for `embed` option value for `cache_has_one` is much safer, forcing the application to specify the value explicitly and will allow us to safely add a new default in a later release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ class Product < ActiveRecord::Base
   has_one   :featured_image
 
   cache_has_many :images
-  cache_has_one :featured_image
+  cache_has_one :featured_image, embed: :id
 end
 
 @product.fetch_featured_image
@@ -191,7 +191,7 @@ Example:
 #### cache_has_many
 
 Options:
-_[:embed]_ When true, specifies that the association should be included with the parent when caching. This means the associated objects will be loaded already when the parent is loaded from the cache and will not need to be fetched on their own. When :ids, only the id of the associated records will be included with the parent when caching.
+_[:embed]_ When true, specifies that the association should be included with the parent when caching. This means the associated objects will be loaded already when the parent is loaded from the cache and will not need to be fetched on their own. When :ids, only the id of the associated records will be included with the parent when caching. Defaults to `:ids`.
 
 _[:inverse_name]_ Specifies the name of parent object used by the association. This is useful for polymorphic associations when the association is often named something different between the parent and child objects.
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -191,7 +191,7 @@ Example:
 #### cache_has_many
 
 Options:
-_[:embed]_ When true, specifies that the association should be included with the parent when caching. This means the associated objects will be loaded already when the parent is loaded from the cache and will not need to be fetched on their own. When :ids, only the id of the associated records will be included with the parent when caching.
+_[:embed]_ When true, specifies that the association should be included with the parent when caching. This means the associated objects will be loaded already when the parent is loaded from the cache and will not need to be fetched on their own. When :ids, only the id of the associated records will be included with the parent when caching. Defaults to `true`.
```
gmcgibbon: The default is `:ids`?

lib/identity_cache/configuration_dsl.rb
```diff
@@ -70,21 +72,22 @@ def cache_has_many(association, embed: :ids, inverse_name: nil)
...
-      #   cache entry as the parent, instead of its own.
+      # * embed: If `true`, IdentityCache will embed the associated record
+      #   in the cache entries for this model, as well as all the embedded
+      #   associations for the associated record recursively.
+      #   If `:id`, it will only embed the id for the associated record.
       # * inverse_name: The name of the parent in the association ( only
       #   necessary if the name is not the lowercase pluralization of the
       #   parent object's class)
-      def cache_has_one(association, embed: true, inverse_name: nil)
+      def cache_has_one(association, embed:, inverse_name: nil)
```
gmcgibbon: This bugs me because it isn't consistent, but I'm fine with fixing that in another release as you mention. 👍

README.md
dylanahsmith: Oops.  Thanks for catching that.  Pushed a fix

PR #439, opened 2020-02-11
Remove inverse_name option. Specify inverse_of on the Active Record association instead.

dylanahsmith:
## Problem

The `inverse_name` option to the `cache_has_*` methods is redundant now that the name can be specified on the active record association.  I think we added it back before that could be done.  After adding the ability to infer the inverse association from the active record association in https://github.com/Shopify/identity_cache/pull/223, I had updated Shopify to remove the use of `inverse_name`, but I think I had only left it in Identity Cache for backwards compatibility reasons.  I probably should have deprecated it at that point.

## Solution

Remove the `inverse_name` cache association option before making the 1.0 release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,24 +144,21 @@ With this code, on cache miss, the product and its associated images will be loa
 
 ### Caching Polymorphic Associations
 
-IdentityCache tries to figure out both sides of an association whenever it can so it can set those up when rebuilding the object from the cache. In some cases this is hard to determine so you can tell IdentityCache what the association should be. This is most often the case when embedding polymorphic associations. The `inverse_name` option on `cache_has_many` and `cache_has_one` lets you specify the inverse name of the association.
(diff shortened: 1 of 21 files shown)


README.md
```diff
@@ -144,25 +144,22 @@ With this code, on cache miss, the product and its associated images will be loa
...
-  belongs_to :owner, :polymorphic => true
+  belongs_to :owner, polymorphic: true
   cache_belongs_to :owner
 end
 
 class Product < ActiveRecord::Base
   include IdentityCache
-  has_many :metafields, :as => 'owner'
-  cache_has_many :metafields, :inverse_name => :owner
+  has_many :metafields, as: 'owner'
```
gmcgibbon: While we're here, we could change this:
```suggestion
  has_many :metafields, as: :owner
```
Is there test coverage for this? If there is, I think we should assert the error message in it.

dylanahsmith: You mean tests for when the inverse isn't inferred.  Yes, they assert that `IdentityCache::InverseAssociationError` is raised in that case.  I'll add an assertion on the error message.

PR #449, opened 2020-04-03
Fix fetch has_many id embedded association after adding to it

dylanahsmith:
cc @mgingras

## Problem

The problem is shown well from the added regression test on a `cache_has_many` `embed: :ids` association

```ruby
    item.associated_records.create!(name: 'foo')
    fetched_associated_records = item.fetch_associated_records
    assert_equal(item.associated_records.length, fetched_associated_records.length)
```

which results in the failure

```
  1) Failure:
NormalizedHasManyTest#test_fetch_association_after_adding_to_it [test/normalized_has_many_test.rb:252]:
Expected: 3
  Actual: 2
```

without the included fix to the code under test.

## Solution

I looked at how Active Record distinguishes between loaded associations and unloaded associations with records added to it and found it checks for this with [`elsif !target.empty?`](https://github.com/rails/rails/blob/v6.0.2.2/activerecord/lib/active_record/associations/collection_association.rb#L48).  I decided to use a similar check, although used `blank?` so that we don't assume that `target` will continue to be initialized to `[]` for an unloaded association.

### DIFF
--- a/test/normalized_has_many_test.rb
+++ b/test/normalized_has_many_test.rb
@@ -244,4 +244,11 @@ class NormalizedHasManyTest < IdentityCache::TestCase
       end
     end
   end
+
+  def test_fetch_association_after_adding_to_it
+    item = Item.fetch(@record.id)
+    item.associated_records.create!(name: 'foo')
+    fetched_associated_records = item.fetch_associated_records
+    assert_equal(item.associated_records.length, fetched_associated_records.length)
+  end
 end
--- a/test/prefetch_associations_test.rb
+++ b/test/prefetch_associations_test.rb
@@ -134,7 +134,7 @@ module IdentityCache
 
       setup_has_many_children_and_grandchildren(@bob)
 
-      associated_records = @bob.associated_records
+      associated_records = Item.find(@bob.id).associated_records
 
       assert_queries(1) do
         assert_memcache_operations(1) do
--- a/dev.yml
+++ b/dev.yml
@@ -3,7 +3,7 @@ name: identity-cache
 up:
   - homebrew:
     - postgresql
-  - ruby: 2.4.1
+  - ruby: 2.4.10
   - railgun
   - bundler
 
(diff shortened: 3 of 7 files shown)


test/normalized_has_many_test.rb
```diff
@@ -244,4 +244,11 @@ def test_respects_should_use_cache_on_association
       end
     end
   end
+
+  def test_fetch_association_after_adding_to_it
```
mgingras, line 248: This test seems to fail for `Item.cache_has_many(:associated_records, embed: true)`, is that expected?

dylanahsmith: Good catch!

I was thinking it was only an issue with id embedding, but I think that is just because the symptoms of the problem were different for id embedding.

I've pushed a commit fixing this for `cache_has_many` `embed: true`

test/prefetch_associations_test.rb
```diff
@@ -134,7 +134,7 @@ def test_prefetch_associations_on_association
 
       setup_has_many_children_and_grandchildren(@bob)
 
-      associated_records = @bob.associated_records
```
dylanahsmith, line 137: This test was failing because `setup_has_many_children_and_grandchildren` was creating records on an association without loading it, which messes with preloading checks.   So calling `prefetch_associations` on records with a modified unloaded association like this may not be ideal, although I think a bigger refactor is needed to handle that complexity well.  For instance, our check to see if an association has already been loaded/fetched assumes that it can just check the first record, but I don't think we want to make a similar assumption for associations being modified which are less likely to be homogeneous.  I would also like to avoid adding too much overhead for the `fetch_includes` fetch option, which can make more assumptions of the state of the records since they aren't coming from application code.

dylanahsmith: Opened https://github.com/Shopify/identity_cache/issues/453 to keep track of this issue.

PR #451, opened 2020-04-06
[Prototype] Object tracking and observability

kirs:
I'm not sure we need this yet, but I wanted to throw a prototype that I've played with that might be useful for anyone else who's going to work on this area.

As mentioned in [caveats](https://github.com/Shopify/identity_cache/blob/master/CAVEATS.md), large blobs and too many embedded associations are a major problem.

At the same time we don't know how often which embedded associations are accessed and if some of them do more harm than good.

This PR brings "object tracking" to IDC. Here's how it works:

1. At the beginning of the request, allocate an array for all objects that will be loaded by IDC
2. Track every read of the embedded association in that array
3. At the end of the request, iterate over objects that were loaded from the IDC and check what embedded associations were accessed.

From the memory/GC point, it means those records would have to be kept around for longer, until the request if finished. That's not great, but I believe is tolerable for a percentage of request.

By subscribing to AS::N events, you could either observe it locally or trace it in production:

```ruby
ActiveSupport::Notifications.subscribe('object_track.identity_cache') do |_event, _, _, _, payload|
  # for local visibility
  # puts "--- #{payload[:object].class.to_s}(id=#{payload[:object].id}): #{payload[:accessed_associations].to_a.join(", ")}"
  # payload[:caller].each do |line|
  #   puts " #{line}"
  # end

  # in production, assuming that Tracing.trace is an abstract way to emit spans
  Tracing.trace('idc.object_tracking', tags: payload) { }
end
```

If your tracing environment is set up to emit data to something like BigQuery (which we use at Shopify), this could be a ground to make a data-driven decision about removing some of embedded associations.

Or maybe I'm overthinking this and there's an easier way?

cc @dylanahsmith @hkdsun @edward

### DIFF
--- a/lib/identity_cache/tracking.rb
+++ b/lib/identity_cache/tracking.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+module IdentityCache
+  module Tracking
+    TrackedObject = Struct.new(:object, :includes, :caller, :accessed_associations)
+
+    extend self
+
+    def tracked_objects
+      Thread.current[:idc_tracked_objects] ||= {}
+    end
+
+    def reset_tracked_objects
+      Thread.current[:idc_tracked_objects] = {}
+    end
+
+    def track_object(object, includes)
+      return unless object_tracking_enabled
+      locations = caller(1, 20)
+      tracked_objects[object] = TrackedObject.new(object, Array(includes), locations, Set.new)
+    end
+
+    def track_association_accessed(object, association_name)
+      return unless object_tracking_enabled
+      obj = self.tracked_objects[object]
+      obj.accessed_associations << association_name if obj
+    end
+
+    def instrument_and_reset_tracked_objects
+      tracked_objects.each do |_, to|
+        ActiveSupport::Notifications.instrument('object_track.identity_cache', {
+          object: to.object,
+          accessed_associations: to.accessed_associations,
+          caller: to.caller
+        })
+      end
+      reset_tracked_objects
+    end
+
+    def with_object_tracking_and_instrumentation
+      begin
+        with_object_tracking { yield }
+      ensure
+        instrument_and_reset_tracked_objects
+      end
+    end
+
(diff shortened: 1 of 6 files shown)


lib/identity_cache/tracking.rb
```diff
@@ -0,0 +1,70 @@
...
+    def track_object(object, includes)
+      return unless object_tracking_enabled
+      locations = caller(1, 20)
+      tracked_objects[object] = TrackedObject.new(object, Array(includes), locations, Set.new)
+    end
+
+    def track_association_accessed(object, association_name)
+      return unless object_tracking_enabled
+      obj = self.tracked_objects[object]
+      obj.accessed_associations << association_name if obj
```
dylanahsmith, line 26: If an association is accessed multiple times, won't this make the array much larger.  Seems like this could result in memory usage and CPU overhead issues.

dylanahsmith: oh, this one is a set, rather it is the `includes` that is the array.

lib/identity_cache/tracking.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    def reset_tracked_objects
+      Thread.current[:idc_tracked_objects] = {}
+    end
+
+    def track_object(object, includes)
+      return unless object_tracking_enabled
+      locations = caller(1, 20)
+      tracked_objects[object] = TrackedObject.new(object, Array(includes), locations, Set.new)
```
dylanahsmith, line 20: Object deserialization is already a hotspot.  So this could add a lot of instrumentation overhead.  I'm guessing the plan is to only use this for a sampling of requests, right?

kirs: Yeah, I think it's only feasible to use this on a tiny percentage of requests.

dylanahsmith: Looks like the PR description already answered my question.

PR #454, opened 2020-04-06
Fix should_use_cache check to avoid calling it on the wrong class.

dylanahsmith:
## Problem

We should try to avoid making assumptions that all the records are of the same class in the association prefetcher, since that may not be the case with single table inheritance.  However, in one place we were still doing `records.first.class`.

## Solution

Use the class that is already passed into the method that was calling `records.first.class`.  This way we avoid an edge case that could otherwise cause a problem.


dylanahsmith:
ping for review

PR #458, opened 2020-04-30
[WIP] Add keys to fetch and fetch_multi instrumentation

Scalvando:
Adding keys to the instrumentation for `fetch` and `fetch_multi` so that you know what keys are having success without looking into hydration events.

### DIFF
--- a/lib/identity_cache/memoized_cache_proxy.rb
+++ b/lib/identity_cache/memoized_cache_proxy.rb
@@ -60,30 +60,35 @@ module IdentityCache
     end
 
     def fetch(key)
-      memo_misses = 0
-      cache_misses = 0
+      memo_miss_keys = EMPTY_ARRAY
+      cache_miss_keys = EMPTY_ARRAY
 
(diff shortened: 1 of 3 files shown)


kirs:
Do you care about the actual full cache key or just about the model name? Is it's the model name, passing `self.class.name` (most of IDC's code runs on the model's class) to instrumentation might be much easier than passing cache keys. It's also much less data to pass around. I imagine that emitting the model name to StatsD would be a lot more acceptable cardinality-wise than the full key.

Scalvando:
> Do you care about the actual full cache key or just about the model name? Is it's the model name, passing `self.class.name` (most of IDC's code runs on the model's class) to instrumentation might be much easier than passing cache keys. It's also much less data to pass around. I imagine that emitting the model name to StatsD would be a lot more acceptable cardinality-wise than the full key.

Ahh I didn't know we would have access to the model name from there. I was going to offload the model parsing to the subscribers.

Scalvando:
Closing this since we only really need to worry about the metrics, just whether or not there was a hit/miss and what the model was. The `{de}hydration` events should be enough to cover this.

PR #459, opened 2020-05-04
Stop passing the column to Connection#type_cast

casperisfine:
That's deprecated since https://github.com/rails/rails/pull/39106

### DIFF
--- a/lib/identity_cache/cached/primary_index.rb
+++ b/lib/identity_cache/cached/primary_index.rb
@@ -60,7 +60,6 @@ module IdentityCache
       def load_multi_from_db(ids)
         return {} if ids.empty?
 
-        ids = ids.map { |id| model.connection.type_cast(id, id_column) }
         records = build_query(ids).to_a
         model.send(:setup_embedded_associations_on_miss, records)
         records.index_by(&:id)
--- a/test/fetch_multi_test.rb
+++ b/test/fetch_multi_test.rb
@@ -211,14 +211,6 @@ class FetchMultiTest < IdentityCache::TestCase
     assert_equal([@joe, @bob, @joe], Item.fetch_multi(@joe.id, @bob.id, @joe.id))
   end
 
-  def test_load_multi_from_db_coerces_ids_to_primary_key_type
-    mock_relation = mock("ActiveRecord::Relation")
-    Item.expects(:where).returns(mock_relation)
-    mock_relation.expects(:includes).returns(stub(to_a: [@bob, @joe, @fred]))
-
-    Item.cached_primary_index.send(:load_multi_from_db, [@bob, @joe, @fred].map(&:id).map(&:to_s))
-  end
-
   def test_fetch_multi_doesnt_freeze_keys
     cache_response = {}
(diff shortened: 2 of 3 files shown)


lib/identity_cache/cached/primary_index.rb
```diff
@@ -57,13 +57,24 @@ def load_one_from_db(id)
...
-
-        ids = ids.map { |id| model.connection.type_cast(id, id_column) }
-        records = build_query(ids).to_a
-        model.send(:setup_embedded_associations_on_miss, records)
-        records.index_by(&:id)
+      if ActiveRecord.gem_version >= Gem::Version.new('6.1')
+        def load_multi_from_db(ids)
+          return {} if ids.empty?
+
+          ids = ids.map { |id| model.connection.type_cast(id) }
```
dylanahsmith, line 64: Do you happen to know why this line is needed in the first place?  It seems like `build_query` on the following line should already be doing casting through `model.where(model.primary_key => id_or_ids)`.

casperisfine: That's a good question, I've been a bit head down and haven't paid too much attention. I'll try it out and see.

dylanahsmith:
Looks like there may be an underlying bug in prefetching associations for polymorphic cache_belongs_to

casperisfine:
Might just be some inaccurate tests, because testing this branch against shopify, seem to work well.

casperisfine:
What's weird is that in these failing tests the `type_cast` is more or less a noop. the `ids` are ints already `[1, 2, 3]`. So there must be some side effects I'm missing.

casperisfine:
Actually scratch that. Sometimes some of the ids are not integers but `#<PolymorphicRecord id: 2, owner_type: "Item", owner_id: 4, created_at: "2020-05-05 14:35:40", updated_at: "2020-05-05 14:35:40">`.

It also makes for fairly suspicious cache keys: `"IDC:8:blob:Item:3231970521778960149:[4, #<PolymorphicRecord id: 2, owner_type: \"Item\", owner_id: 4, created_at: \"2020-05-05 14:35:40\", updated_at: \"2020-05-05 14:35:40\">]"`

dylanahsmith:
Yeah, that's what I mean about it looking like a bug.  I'll open another PR with a fix.

lib/identity_cache/cached/primary_index.rb
```diff
@@ -60,7 +60,6 @@ def load_one_from_db(id)
       def load_multi_from_db(ids)
         return {} if ids.empty?
 
-        ids = ids.map { |id| model.connection.type_cast(id, id_column) }
```
dylanahsmith, line 63: I tried running the tests `puts caller if ids != ids.map { |id| model.connection.type_cast(id, id_column) }` to see if we were failing to coerce the ids before building the cache keys, since that would cause more problems.  In the process, I found a couple of tests for this line, but they were just calling this private method directly with heavy mocking/stubbing, so they don't represent real usage.  I'm pushing a commit to remove those useless tests.

PR #463, opened 2020-06-12
Only expire the cache when the record has changed, or it is forced

DougEdey:
We discovered that `after_commit` is part of the ActiveRecord object lifecycle, not the database, so it's called after the commit method, not after the persistance layer has commited changed data

To accomodate this, I'm using the [`ar_transaction_changes`](https://github.com/dylanahsmith/ar_transaction_changes) gem to see if there are any records, or the object was destroyed to clear the cached attributes.

Additionally, I feel like `expire_cache` should be lazy, and default to this behaviour, but still give the ability to force it for some reason 

I'm willing to bet this is being used in a lot of places, but for no reason, infact, I know it's used with `update_all` batch commits in Shopify as callbacks aren't used, but because the default is `force: false` this should hopefully improve some behaviour here.

A lot of the tests relied on this behaviour, so I broke a fair number and fixed them by changing the values.

### DIFF
--- a/test/attribute_cache_test.rb
+++ b/test/attribute_cache_test.rb
@@ -39,13 +39,13 @@ class AttributeCacheTest < IdentityCache::TestCase
     end
   end
 
-  def test_cached_attribute_values_are_expired_from_the_cache_when_an_existing_record_is_saved
+  def test_cached_attribute_values_are_not_expired_from_the_cache_when_an_existing_record_is_saved_with_no_changes
     assert_queries(1) { assert_equal 'foo', AssociatedRecord.fetch_name_by_id(1) }
     assert_queries(0) { assert_equal 'foo', AssociatedRecord.fetch_name_by_id(1) }
-
+    
     @record.save!
 
-    assert_queries(1) { assert_equal 'foo', AssociatedRecord.fetch_name_by_id(1) }
+    assert_queries(0) { assert_equal 'foo', AssociatedRecord.fetch_name_by_id(1) }
   end
 
   def test_cached_attribute_values_are_expired_from_the_cache_when_an_existing_record_with_changed_attributes_is_saved
--- a/test/save_test.rb
+++ b/test/save_test.rb
@@ -41,6 +41,15 @@ class SaveTest < IdentityCache::TestCase
     @record.save
   end
 
+  def test_update_no_change
+    # Regular flow, write index id, write index id/tile, delete data blob since Record has changed
+    expect_cache_delete("#{NAMESPACE}attr:Item:id:id/title:#{cache_hash('"1"/"bob"')}").never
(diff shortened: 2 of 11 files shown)


test/attribute_cache_test.rb
```diff
@@ -39,13 +39,13 @@ def test_nil_is_stored_in_the_cache_on_cache_misses
     end
   end
 
-  def test_cached_attribute_values_are_expired_from_the_cache_when_an_existing_record_is_saved
+  def test_cached_attribute_values_are_not_expired_from_the_cache_when_an_existing_record_is_saved_with_no_changes
```
dylanahsmith, line 42: This is changing the purpose of this test.  Unless the test was redundant, we should be adding a new test for the new behaviour and just update this one to test the behaviour it was intended for.  This can be done by updating some attribute instead of calling `save!` without changing any attributes.

test/save_test.rb
```diff
@@ -41,6 +41,15 @@ def test_update
     @record.save
   end
 
+  def test_update_no_change
+    # Regular flow, write index id, write index id/tile, delete data blob since Record has changed
+    expect_cache_delete("#{NAMESPACE}attr:Item:id:id/title:#{cache_hash('"1"/"bob"')}").never
+    expect_cache_delete("#{NAMESPACE}attr:Item:id:title:#{cache_hash('"bob"')}").never
+    expect_cache_delete(@blob_key).never
```
dylanahsmith, line 48: This type of test is quite fragile.  Not only is it coupled to the internals of the library, but also changes to those internals could cause this test to no longer test anything relevant by it instead deleting slightly different cache key.

A way to test this with less internal coupling would be to do a fetch surrounded by `assert_no_queries` to make sure the cache wasn't expired.

test/save_test.rb
```diff
@@ -41,6 +41,15 @@ def test_update
...
+
+    @record.save
```
dylanahsmith, line 50: If the return value for the `save` isn't going to be checked, then use `save!` instead to make sure the test fails in an easy to debug way if there are unexpected validation errors.  You can also make this change when refactoring existing tests that have this anti-pattern.

test/save_test.rb
```diff
@@ -69,11 +78,21 @@ def test_touch_will_expire_the_caches
...
   end
 
-  def test_expire_cache_works_in_a_transaction
+  def test_expire_cache_works_in_a_transaction_when_forced
     expect_cache_delete("#{NAMESPACE}attr:Item:id:id/title:#{cache_hash('"1"/"bob"')}")
     expect_cache_delete("#{NAMESPACE}attr:Item:id:title:#{cache_hash('"bob"')}")
     expect_cache_delete(@blob_key)
 
+    ActiveRecord::Base.transaction do
+      @record.expire_cache(force: true)
```
dylanahsmith, line 87: The fact that this test needs to be updated in this way shows the breaking change I mentioned in the change to `expire_cache`

PR #464, opened 2020-06-12
Skip expiring record cache on save with no db update

dylanahsmith:
## Problem

When `ActiveRecord::Base#save!` is called without modifying any attributes, then active record will skip the database update, however, it will still result in the record being expired in identity cache.  This happens because active record still calls the after_commit callback in this case.

## Solution

Make the cache expiry on update to be conditional on `transaction_changed_attributes.present?` so that it will skip these unnecessary cache expiries.


test/attribute_cache_test.rb
```diff
@@ -43,7 +43,7 @@ def test_cached_attribute_values_are_expired_from_the_cache_when_an_existing_rec
     assert_queries(1) { assert_equal 'foo', AssociatedRecord.fetch_name_by_id(1) }
     assert_queries(0) { assert_equal 'foo', AssociatedRecord.fetch_name_by_id(1) }
 
-    @record.save!
+    @record.update!(updated_at: @record.updated_at + 1)
```
DougEdey, line 46: Would a `touch` be easier to understand here/cleaner?

dylanahsmith: I wanted to avoid having the test fail because the record's `updated_at` matched the current time, in which case we are trying to skip expiring the cache.

DougEdey: ah right, the timestamp isn't precise enough, makes sense

PR #470, opened 2020-10-02
Autocorrect rubocop offenses

dylanahsmith:
## Problem

I was going to open another pull request, but noticed that I got rubocop errors on CI (which I didn't see locally).  It looks like these got introduced in a newer rubocop version.

## Solution

Use rubocop autocorrect


camilo (approved):
Since CI passes.

PR #471, opened 2020-10-02
Expire caches before other after_commit callbacks

dylanahsmith:
Green CI depends on rubocop autocorrections in https://github.com/Shopify/identity_cache/pull/470 

## Problem

We noticed a race condition between identity cache's cache expiration after_commit callback and background jobs enqueued form another after_commit callback for the same record.  This can result in stale data being loaded from the cache if the after_commit callback that enqueues the job is called after the identity cache's cache expiration after_commit callback.

The order that after_commit callbacks are called is the opposite order from when they are defined by calling `after_commit`.  By convention we normally include modules near the top of a model and Identity Cache's `after_commit` callback is defined when the `IdentityCache`/`IdentityCache::WithoutPrimaryIndex` is included, so typically identity cache's cache expiration is done after application defined after_commit callbacks.

## Solution

The ar_transaction_changes had a similar concern with ordering of callbacks (although it wanted to do something **after** all other after_commit callbacks), which it solved by [wrapping the _run_commit_callbacks method](https://github.com/dylanahsmith/ar_transaction_changes/blob/v1.1.7/lib/ar_transaction_changes.rb) that is used to call the after_commit callbacks.  This is a method defined by ActiveSupport::Callbacks's `define_callbacks` class method.

I took a similar approach here, wrapping `_run_commit_callbacks` and expiring caches before calling `super` in that override.

I've also included a regression test that tries to fetch the updated record in another after_commit callback and fails without the corresponding fix.

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -3,12 +3,6 @@ module IdentityCache
   module QueryAPI
     extend ActiveSupport::Concern
 
-    included do |base|
-      base.after_commit do
-        expire_cache if destroyed? || transaction_changed_attributes.present?
-      end
-    end
-
     module ClassMethods
       # Prefetches cached associations on a collection of records
       def prefetch_associations(includes, records)
@@ -152,6 +146,25 @@ module IdentityCache
 
         associations_for_identity_cache.compact
       end
+    end
+
+    included do |base|
+      # Make sure there is at least once after_commit callback so that _run_commit_callbacks
+      # is called, which is overridden to do an early after_commit callback
+      base.after_commit {}
+    end
+
+    # Override the method that is used to call after_commit callbacks so that we can
+    # expire the caches before other after_commit callbacks. This way we can avoid stale
+    # cache reads that happen from the ordering of callbacks. For example, if an after_commit
+    # callback enqueues a background job, then we don't want it to be possible for the
+    # background job to run and load data from the cache before it is invalidated.
+    def _run_commit_callbacks
+      if destroyed? || transaction_changed_attributes.present?
+        expire_cache
+        expire_parent_caches
+      end
+      super
     end
 
     # Invalidate the cache data associated with the record.
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -148,10 +148,11 @@ module IdentityCache
       end
     end
 
+    no_op_callback = proc {}
     included do |base|
       # Make sure there is at least once after_commit callback so that _run_commit_callbacks
       # is called, which is overridden to do an early after_commit callback
-      base.after_commit {}
+      base.after_commit(&no_op_callback)
     end
 
     # Override the method that is used to call after_commit callbacks so that we can
--- a/lib/identity_cache/parent_model_expiration.rb
+++ b/lib/identity_cache/parent_model_expiration.rb
@@ -37,9 +37,6 @@ module IdentityCache
     included do
       class_attribute(:parent_expiration_entries)
       self.parent_expiration_entries = Hash.new { |hash, key| hash[key] = [] }
-      after_commit do
-        expire_parent_caches if destroyed? || transaction_changed_attributes.present?
-      end
     end
 
     def expire_parent_caches
--- a/test/cache_invalidation_test.rb
+++ b/test/cache_invalidation_test.rb
@@ -182,4 +182,17 @@ class CacheInvalidationTest < IdentityCache::TestCase
       @record.class.fetch(@record.id)
     end
   end
+
+  def test_cache_expiry_before_other_after_commit_callbacks
+    Item.fetch(@record.id) # fill cache
+
+    record_id = @record.id
+    after_commit_fetched = nil
+    Item.after_commit do
+      after_commit_fetched = Item.fetch(record_id)
+    end
+    @record.update!(updated_at: @record.updated_at + 1)
+
+    assert_equal(after_commit_fetched.updated_at, @record.updated_at)
+  end
 end
--- a/test/cache_invalidation_test.rb
+++ b/test/cache_invalidation_test.rb
@@ -195,4 +195,19 @@ class CacheInvalidationTest < IdentityCache::TestCase
 
     assert_equal(after_commit_fetched.updated_at, @record.updated_at)
   end
+
+  def test_cache_not_expired_until_after_transaction
+    log = []
+    subscribe_to_sql_queries(->(sql) { log << [:sql, sql] }) do
+      subscribe_to_cache_operations(->(op) { log << [:cache, op] }) do
+        @record.update!(title: 'foo2')
+      end
+    end
+    assert_equal([:sql, :sql, :sql, :cache], log.map(&:first))
+    sql_queries = log.first(3).map(&:last)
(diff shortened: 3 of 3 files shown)


camilo:
cc @mikedasilva this might have to do with the issue you were facing (remember this is a public repo)

lib/identity_cache/query_api.rb
```diff
@@ -154,6 +148,25 @@ def cache_fetch_includes
       end
     end
 
+    included do |base|
+      # Make sure there is at least once after_commit callback so that _run_commit_callbacks
+      # is called, which is overridden to do an early after_commit callback
+      base.after_commit {}
```
camilo, line 154: interesting, I wonder what kind of overhead we are incurring here? if we had 100s of models we'd be adding 100s of callbacks; while they are NOOP we probably still have allocate many objects for no reason.

This onto itself is not a reason not to ship, but would be good to remember to measure that somehow when we bump the version of the library on a reasonably large app.

lib/identity_cache/query_api.rb
```diff
@@ -154,6 +148,25 @@ def cache_fetch_includes
...
+    end
+
+    # Override the method that is used to call after_commit callbacks so that we can
+    # expire the caches before other after_commit callbacks. This way we can avoid stale
+    # cache reads that happen from the ordering of callbacks. For example, if an after_commit
+    # callback enqueues a background job, then we don't want it to be possible for the
+    # background job to run and load data from the cache before it is invalidated.
+    def _run_commit_callbacks
```
camilo, line 162: I'm assuming this is not quite a public RoR API. I worry about weird interactions with the main implementation, I guess we are just calling `super`. Maybe I'm just being paranoid .

lib/identity_cache/query_api.rb
```diff
@@ -154,6 +148,25 @@ def cache_fetch_includes
...
+      if destroyed? || transaction_changed_attributes.present?
+        expire_cache
+        expire_parent_caches
+      end
```
camilo, line 166: nit: this could be factored into a function `identity_cache_expire_caches` only reason I'm suggesting that is that would make it more clear in a stack trace what is going on. Take it or leave it not a strong feeling here.

camilo (commented):
This should work;  I wonder if we are losing some behaviour we used to expect by removing the expiration from the callback stack I can't think of anything but I'm sure there is logic that is executed around each callback and I'm not 100% confident our tests would capture that.

lib/identity_cache/query_api.rb
dylanahsmith: We were previously doing this twice per model, now it is done only once, so I don't think it is that significant.  I could switch to using a no-op method or re-using a proc to avoid creating a proc per-model.

lib/identity_cache/query_api.rb
dylanahsmith: Looks like the ActiveSupport::Callbacks documentation says to call `run_callbacks` and it looks like [`_run_#{name}_callbacks` is just a wrapper around run_callbacks](https://github.com/rails/rails/blob/v6.0.3.3/activesupport/lib/active_support/callbacks.rb#L824-L826).  So it looks like `run_callbacks` would be the more public and stable API to override.  However, it is also probably an implementation detail that Active Record callbacks are implemented using `ActiveSupport::Callbacks` given their use of wrapper methods rather suggesting that `set_callback` be used directly to define these callbacks.

dylanahsmith:
> cc @mikedasilva this might have to do with the issue you were facing (remember this is a public repo)

This was directly in respond to the issue you brought to my attention.

> I wonder if we are losing some behaviour we used to expect by removing the expiration from the callback stack I can't think of anything

_run_commit_callbacks is defined by `define_callbacks` (from ActiveSupport::Callbacks), so it would only include callback code.

> but I'm sure there is logic that is executed around each callback and I'm not 100% confident our tests would capture that.

Looks like we are definitely missing some test coverage to make sure expiration happens after commit rather than before commit.  For instance, the test suite passed with the following change

```diff
diff --git a/lib/identity_cache/query_api.rb b/lib/identity_cache/query_api.rb
index f8e03d2..b5a0615 100644
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -151,7 +151,7 @@ module IdentityCache
     included do |base|
       # Make sure there is at least once after_commit callback so that _run_commit_callbacks
       # is called, which is overridden to do an early after_commit callback
-      base.after_commit {}
+      base.before_commit {}
     end
 
     # Override the method that is used to call after_commit callbacks so that we can
@@ -159,7 +159,7 @@ module IdentityCache
     # cache reads that happen from the ordering of callbacks. For example, if an after_commit
     # callback enqueues a background job, then we don't want it to be possible for the
     # background job to run and load data from the cache before it is invalidated.
-    def _run_commit_callbacks
+    def _run_before_commit_callbacks
       if destroyed? || transaction_changed_attributes.present?
         expire_cache
         expire_parent_caches
```

so I'll add a test for that.  Note that `before_commit` is implemented as a separate callback (although is undocumented).

lib/identity_cache/query_api.rb
camilo: you are right, don't mind me

camilo:
>Looks like we are definitely missing some test coverage to make sure expiration happens after commit rather than before commit. For instance, the test suite passed with the following change


+1 . I resolved the wrong comment by mistake but I agree with this.

mikedasilva (approved):
Thanks for fixing this issue!

lib/identity_cache/query_api.rb
```diff
@@ -154,6 +148,26 @@ def cache_fetch_includes
       end
     end
 
+    no_op_callback = proc {}
+    included do |base|
+      # Make sure there is at least once after_commit callback so that _run_commit_callbacks
+      # is called, which is overridden to do an early after_commit callback
+      base.after_commit(&no_op_callback)
```
gmcgibbon, line 155: Why is this necessary? I believe `_run_commit_callbacks` gets called regardless of the number of callbacks.

dylanahsmith: Looks like it isn't needed on rails master, but is needed for Active Record version 6.0.3.3.  I added this in response to seeing test failures.  There is some conditional logic around the call to `_run_commit_callbacks`, but I hadn't dug into why exactly this is needed.

gmcgibbon:
```suggestion
      after_commit(&proc {}) if _commit_callbacks.none?
```
You're right, we need this for previous versions we support.

gmcgibbon (approved):

lib/identity_cache/query_api.rb
dylanahsmith: Active Record does in fact check to see if there are commit callbacks in [has_transactional_callbacks?](https://github.com/rails/rails/blob/v6.0.3.3/activerecord/lib/active_record/transactions.rb#L462-L464) which is a method that even exists on rails master.  `with_transaction_returning_status` uses that method in rails 6.0 to conditionally call `add_to_transaction` which is needed to make sure the commit callbacks are called.  On rails master, it unconditionally calls `add_to_transaction`, but uses `has_transactional_callbacks?` to determine the argument to pass into that method (`ensure_finalize`).

PR #474, opened 2020-10-11
[DEV] Fixing double logging of cache hits and misses

glebglazov:
Due to the fact that in Rails logger is extended to output to console and server respectively
https://github.com/rails/rails/blob/4cf7559/activerecord/lib/active_record/railtie.rb#L64
https://github.com/rails/rails/blob/4cf7559/railties/lib/rails/commands/server/server_command.rb#L85
there is a downside of using `IdentityCache.logger.debug` and calling it once again inside that block that results in duplication of logs:
image 
This pull request fixes the problem by using guard-clause instead of wrapping the code in `logger.debug` call making it log only once:
image

### DIFF
--- a/lib/identity_cache/memoized_cache_proxy.rb
+++ b/lib/identity_cache/memoized_cache_proxy.rb
@@ -195,15 +195,15 @@ module IdentityCache
     end
 
     def log_multi_result(keys, memo_miss_keys, cache_miss_keys)
-      IdentityCache.logger.debug do
-        memoized_keys = keys - memo_miss_keys
-        cache_hit_keys = memo_miss_keys - cache_miss_keys
-        missed_keys = cache_miss_keys
+      return unless IdentityCache.logger.level == Logger::DEBUG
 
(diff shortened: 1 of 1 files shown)


lib/identity_cache/memoized_cache_proxy.rb
```diff
@@ -195,15 +195,15 @@ def memoizing?
...
-      IdentityCache.logger.debug do
-        memoized_keys = keys - memo_miss_keys
-        cache_hit_keys = memo_miss_keys - cache_miss_keys
-        missed_keys = cache_miss_keys
-
-        memoized_keys.each { |k| IdentityCache.logger.debug("[IdentityCache] (memoized) cache hit for #{k} (multi)") }
-        cache_hit_keys.each { |k| IdentityCache.logger.debug("[IdentityCache] (backend) cache hit for #{k} (multi)") }
-        missed_keys.each { |k| IdentityCache.logger.debug("[IdentityCache] cache miss for #{k} (multi)") }
-      end
+      return unless IdentityCache.logger.level == Logger::DEBUG
```
dylanahsmith, line 198:
```suggestion
      return unless IdentityCache.logger.debug?
```

PR #476, opened 2020-12-07
Recursively install parent expiry hooks when expirying parent caches

dylanahsmith:
## Problem

We had a problem in development with records not being expired when they were saved.  @andrewhassan helped me track down this problem, where we found the record was a deeply embedded association and the debug logs showed that the record along with its parent was being expired, but not the grandparent record.  So it looked like the cache expiration didn't happen when fetching the grandparent record from the cache.

I realized this is because we lazily install parent expiration hooks, in order to reduce eager loading of models for boot time performance.  On cache expiration, we install the pending parent expiration hooks, but we weren't doing this recursively.

## Solution

This PR moves the call to `ParentModelExpiration.install_pending_parent_expiry_hooks` just before the call to `parent_expiration_entries` that it affects, in a method that gets called recursively during parent expiration.

Since `ParentModelExpiration.install_pending_parent_expiry_hooks` will be called more often, I reduced the `model.name.demodulize` overhead from calling this method using a fast path for when all hooks have been installed.

I've also included a regression test which fails without the corresponding fix.  This situation wasn't well tested because it requires inserting fixtures and filling the cache before setting up the models, which I simulated by tearing down and setting up the models again in the test, similar to what we do in SchemaChangeTest#test_embed_existing_cache_has_many

### DIFF
--- a/lib/identity_cache/parent_model_expiration.rb
+++ b/lib/identity_cache/parent_model_expiration.rb
@@ -21,6 +21,7 @@ module IdentityCache
       end
 
       def install_pending_parent_expiry_hooks(model)
+        return if lazy_hooks.empty?
         name = model.name.demodulize
         if (hooks = lazy_hooks.delete(name))
           hooks.each(&:install)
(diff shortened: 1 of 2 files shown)


gmcgibbon (approved):
The parent model expiration concern is a little complex, but from what I can gather:
- The recursion here is necessary because we may not know the full association parent chain due to autoloading
- Our previous recursive cache expire test was unrealistic because we manually install the hook
- `return if lazy_hooks.empty?` isn't necessarily needed, but is a performance optimization

👍 LGTM, thanks for fixing this!

PR #479, opened 2020-12-15
Use Github Actions for CI instead of travis

dylanahsmith:
## Problem

travis-ci.org has severely limited their capacity and will be put into read-only mode at the end of the year

## Solution

We already started using Github Actions for rubocop, so I've extended it to run the tests as well.


.github/workflows/ci.yml
```diff
@@ -1,26 +1,77 @@
...
 jobs:
   build:
-
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        entry:
+          - { ruby: 2.5, gemfile: "Gemfile.min-supported" }
+          - { ruby: 2.7, gemfile: "Gemfile.latest-release", rubocop: true }
```
gmcgibbon, line 13:
```suggestion
          - { ruby: 2.7, gemfile: "Gemfile.latest-release", rubocop: "rubocop" }
```
The check names for rubocop and edge are kind of weird. Would it make sense to make the value a string so the check is more readable?

gmcgibbon (commented):
I was just thinking about this the other day. Thanks for being a better maintainer than me. 😄 

Looks like there's a failure on edge rails related to the preloader.

dylanahsmith:
> Looks like there's a failure on edge rails related to the preloader.

I opened https://github.com/Shopify/identity_cache/pull/478 for that already

.github/workflows/ci.yml
```diff
@@ -1,26 +1,77 @@
 name: CI
 
-on: [push]
+on: [push, pull_request]
```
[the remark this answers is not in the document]
dylanahsmith: They aren't redundant for outside contributions, where the pull request branch is pushed to a fork, since github actions aren't automatically enabled on forks.

.github/workflows/ci.yml
dylanahsmith: I don't understand in what way they are weird.  Do you mean because we don't specify the value for all matrix entries?

Using a string seems misleading, since the value is being used as a boolean.  Using a string seems like it suggests that the string value will be used, rather than it being just used as a truthy value.

.github/workflows/ci.yml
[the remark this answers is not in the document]
dylanahsmith: Maybe we could ignore the event in the case where they are redundant.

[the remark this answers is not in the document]
dylanahsmith: Added the conditional to skip the job if it is for a pull request event for a branch that is on this repo.  You can see that on the checks for this PR, which includes "CI / build (pull_request) Skipped".  I created a fork and opened a PR from that fork to test that scenario (https://github.com/Shopify/identity_cache/pull/480) which used to only show the CLA check with `on: [push]` and now runs CI when the PR is opened or when a push is made to the PR branch.

.github/workflows/ci.yml
gmcgibbon: The name of the job that runs rubocop is `CI / build (2.7, Gemfile.latest-release, true)` and the one that runs edge rails `CI / build (2.7, Gemfile.rails-edge, true)`. The `true` means multiple things, and I'm wondering if a string value would result in `CI / build (2.7, Gemfile.latest-release, rubocop)` and `CI / build (2.7, Gemfile.rails-edge, edge)`.

Looking again, I guess it is only really unclear about rubocop. Feel free to ignore

gmcgibbon (approved):

.github/workflows/ci.yml
dylanahsmith: The `CI` part is the workflow name.  The `build (2.7, Gemfile.latest-release, true)` part is the job name, which seems to be created automatically from github.  We can specify the job names explicitly, so I've pushed a commit to do that.

dylanahsmith: The one downside is that the job name seems to be used without expression substitution when the job is skipped and the name uses the `matrix` context object.

PR #481, opened 2021-01-04
License year 2021

gmcgibbon:
Update license year to 2021.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2013 Shopify
+Copyright (c) 2013-2021 Shopify
 
 MIT License
 


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2013 Shopify
+Copyright (c) 2021 Shopify
```
dylanahsmith, line 1: I think it should be a range to indicate that it includes code that was written earlier than the current year.

```suggestion
Copyright (c) 2013-2021 Shopify
```

gmcgibbon: Done!

dylanahsmith (approved):

PR #484, opened 2021-01-14
Run tests against ruby 3

ionut998:
Add CI entry for Ruby 3.0 - Rails edge
Versions of `memcached_store` prior to `2.1.5` are incompatible with Ruby 3. The fix for using the last argument as keyword parameters was fixed in memcached_store in Shopify/memcached_store#56


dylanahsmith:
Fixed by https://github.com/Shopify/identity_cache/pull/485

PR #485, opened 2021-01-15
Update dependencies used in development and on CI

dylanahsmith:
## Problem

I noticed that we were using some old development dependencies (e.g. mocha 0.14.0 is from 2013), weren't actually testing the latest version of optional runtime dependencies (e.g. memcached_store) and weren't testing the latest version of ruby (3) on CI.

## Solution

I used the gemspec to specify the common development and CI dependencies and pinned these dependencies to a major version.

I've added a version constraint for optional runtime dependencies in gemfiles/Gemfile.min-supported and removed these constraints from gemfiles/Gemfile.latest-release so that we can test old and new versions of these gems.

The above allowed ruby 3 to be tested without issues, so I updated CI to use ruby 3 as the latest ruby version.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,9 @@
 source 'https://rubygems.org'
 gemspec
 
+gem 'rubocop', '~> 1.5'
 gem 'mysql2', '~> 0.5.3'
 gem 'pg', ">= 0.18", "< 2.0"
-gem 'rubocop', '~> 1.5'
+
 gem 'byebug', platform: :mri
+gem 'stackprof', platform: :mri
--- a/Gemfile
+++ b/Gemfile
@@ -3,8 +3,13 @@ source 'https://rubygems.org'
 gemspec
 
 gem 'rubocop', '~> 1.5'
-gem 'mysql2', '~> 0.5.3'
-gem 'pg', ">= 0.18", "< 2.0"
+
+gem 'mysql2', '~> 0.5.3', platform: :mri
+gem 'pg', ">= 0.18", "< 2.0", platform: :mri
+gem 'memcached', '~> 1.8.0', platform: :mri
+gem 'memcached_store', '~> 1.0.0', platform: :mri
+gem 'dalli', '~> 2.7.11'
+gem 'cityhash', '~> 0.6.0', platform: :mri
 
 gem 'byebug', platform: :mri
 gem 'stackprof', platform: :mri
--- a/gemfiles/Gemfile.rails-edge
+++ b/gemfiles/Gemfile.rails-edge
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 gemspec path: '..'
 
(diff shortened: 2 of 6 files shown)


martelogan (approved):
LGTM. Thanks for this 👍

identity_cache.gemspec
```diff
@@ -38,20 +38,8 @@ Gem::Specification.new do |gem|
...
 
-  gem.add_development_dependency('memcached', '~> 1.8.0')
-  gem.add_development_dependency('memcached_store', '~> 1.0.0')
-  gem.add_development_dependency('dalli')
-  gem.add_development_dependency('rake')
-  gem.add_development_dependency('mocha', '0.14.0')
-  gem.add_development_dependency('spy')
-  gem.add_development_dependency('minitest', '>= 2.11.0')
-
-  if RUBY_PLATFORM == 'java'
```
gmcgibbon, line 49: I believe this changes behaviour when someone tries to run this gem on JRuby, right?

Gemfile
```diff
@@ -2,7 +2,14 @@
...
+
+gem 'mysql2', '~> 0.5.3', platform: :mri
+gem 'pg', ">= 0.18", "< 2.0", platform: :mri
+gem 'memcached', '~> 1.8.0', platform: :mri
+gem 'memcached_store', '~> 1.0.0', platform: :mri
+gem 'dalli', '~> 2.7.11'
+gem 'cityhash', '~> 0.6.0', platform: :mri
+
 gem 'byebug', platform: :mri
+gem 'stackprof', platform: :mri
```
gmcgibbon, line 15: Why are we specifying `platform: :mri` in these gems?

gemfiles/Gemfile.rails-edge
```diff
@@ -3,5 +3,8 @@ gemspec path: '..'
 
 gem 'activerecord', github: 'rails/rails'
```
rafaelfranca, line 4:
```suggestion
gem 'activerecord', github: 'rails/rails', branch: 'main'
```

gemfiles/Gemfile.rails-edge
```diff
@@ -3,5 +3,8 @@ gemspec path: '..'
...
 gem 'activesupport', github: 'rails/rails'
```
rafaelfranca, line 5:
```suggestion
gem 'activesupport', github: 'rails/rails', branch: 'main'
```

rafaelfranca (approved):

identity_cache.gemspec
dylanahsmith: It was just raising before.  However, there is no need to add that constraint here, since Identity Cache doesn't itself require MRI ruby, even though we end up testing with MRI specific backends.

Gemfile
dylanahsmith: Yeah, these are native extensions, so we can just not install them when not using MRI.

PR #490, opened 2021-02-19
Using attribute instead of func to enhance compatibility.

xuxiangyang:
Some gem such as [polymorphic_integer_type](https://github.com/clio/polymorphic_integer_type) may overwrite the func

### DIFF
--- a/lib/identity_cache/cached/attribute.rb
+++ b/lib/identity_cache/cached/attribute.rb
@@ -96,7 +96,7 @@ module IdentityCache
       end
 
       def new_cache_key(record)
-        new_key_values = key_fields.map { |field| record.send(field) }
+        new_key_values = key_fields.map { |field| record.attributes[field.to_s] }
         cache_key_from_key_values(new_key_values)
       end
 
@@ -109,7 +109,7 @@ module IdentityCache
           elsif record.persisted? && changes.key?(field_string)
             changes[field_string]
           else
-            record.send(field)
+            record.attributes[field.to_s]
           end
         end
         cache_key_from_key_values(old_key_values)


rafaelfranca:
I don't think this is correct, we want the method exactly because applications might override the attribute readers. Is there a reason why the method defined by this gem you linked can't be used to generate the cache key?

xuxiangyang:
With under classes
```ruby
create_table :feeds do |t|
  t.integer :feedable_type, null: false
  t.integer :feedable_id, null: false
  t.integer :views_count, null: false
  t.index [:feedable_type, :feedable_id], unique: true
end

create_table :videos do |t|
  t.text :cover_url
  t.text :video_url
end

create_table :images do |t|
  t.text :image_url
end

class Feed < ApplicationRecord
  include IdentityCache
  belongs_to :feedable, polymorphic: { 0 => "Video", 1 => "Image" }

  cache_index :feedable_type, :feedable_id, unique: true
end

class Video < ApplicationRecord
  has_one :feed, as: :feedable
end

class Image < ApplicationRecord
  has_one :feed, as: :feedable
end
```

[polymorphic_integer_type](https://github.com/clio/polymorphic_integer_type)  will overwite  `feedable_type` which return `String` instead of `Integer`.  The reason we must use Integer as polymorphic type is that upper application need fetch `feed` with `feedable_type`  and `feedable_id`. using class name as  polymorphic type will be unstable in some case.

Make an example with Feed with Video Type, when IdentityCache generate the cache key, feedable_type will use `Video`  to compute. The cache will always miss with `Feed.fetch_by_feedable_type_and_feedable_id(0, id)`.

I'm not sure whether it's a good idea to change IdentityCache code....

rafaelfranca:
It looks like `polymorphic_integer_type ` should be also adding a method called `original_feedable_type` that would return integer so you could be using it as the cache key.

```
cache_index :original_feedable_type, :feedable_id, unique: true
```

xuxiangyang:
thanks, I will try this definition

Issue #493, opened 2021-05-01
Rails Optimized Cache::Entry coder breaks IDC


jeromecornet:
Looks like this PR https://github.com/rails/rails/pull/42025 introduced a raw: parameter to the deserialize_entry of the memcache used by IDC (when configured to use mem_cache_store)

@byroot I don't really know whether the intention is to update IDC or keep backwards compatibility in rails, so I'm raising the issue against IDC (although it may be a problem with the original PR) but I figured that you would see it anyway.

=> #<ArgumentError: missing keyword: :raw>
Backtrace
```
"/app/vendor/bundle/ruby/2.7.0/bundler/gems/rails-08e3e663ca70/activesupport/lib/active_support/cache/mem_cache_store.rb:247:in `deserialize_entry'",
 "/app/vendor/bundle/ruby/2.7.0/bundler/gems/identity_cache-01e90ed9fffe/lib/identity_cache/mem_cache_store_cas.rb:14:in `block (4 levels) in cas'",
 "/app/vendor/bundle/ruby/2.7.0/gems/dalli-2.7.11/lib/dalli/client.rb:276:in `cas_core'",
 "/app/vendor/bundle/ruby/2.7.0/gems/dalli-2.7.11/lib/dalli/client.rb:116:in `cas'",
 "/app/vendor/bundle/ruby/2.7.0/bundler/gems/identity_cache-01e90ed9fffe/lib/identity_cache/mem_cache_store_cas.rb:13:in `block (3 levels) in cas'",
 "/app/vendor/bundle/ruby/2.7.0/gems/dalli-2.7.11/lib/dalli/client.rb:267:in `with'",
 "/app/vendor/bundle/ruby/2.7.0/bundler/gems/identity_cache-01e90ed9fffe/lib/identity_cache/mem_cache_store_cas.rb:12:in `block (2 levels) in cas'",
 "/app/vendor/bundle/ruby/2.7.0/bundler/gems/rails-08e3e663ca70/activesupport/lib/active_support/cache.rb:775:in `block in instrument'",
 "/app/vendor/bundle/ruby/2.7.0/bundler/gems/rails-08e3e663ca70/activesupport/lib/active_support/notifications.rb:207:in `block in instrument'",
 "/app/vendor/bundle/ruby/2.7.0/bundler/gems/rails-08e3e663ca70/activesupport/lib/active_support/notifications/instrumenter.rb:44:in `instrument'",
[...]

PR #494, opened 2021-05-03
Fix compatibility with Rails edge

casperisfine:
Fix: https://github.com/Shopify/identity_cache/issues/493

https://github.com/rails/rails/pull/42025 changed the signature of `serialize_entry`. cc @jeromecornet


dylanahsmith (approved):

dylanahsmith:
Also, nice work on the upstream changes!

PR #495, opened 2021-06-07
Add expire methods by key values

Hectorhammett:
There are situations where performance is crucial. For those reasons it can be desired to utilize methods that do not instantiate active record objects, for example `pluck` for reading data and `upsert_all` for writing data. Utilizing these gives the developer the flexibility to read or update the needed information without needing to take the time to instantiate a new Active Record instance for each of the objects as this instantiation can really add up on processing time.

Following the `pluck` example utilizing the IDC feature of `cache_attribute` it is super useful in the case that we need just one value of one record. In the case where we need to update these records it might be preferred to utilize `upsert_all` which does not instantiate any ActiveRecords and we don't have a way to expire the cache. A possible way to accomplish this is to fetch the data before doing the update or update them through the ActiveRecord object, although the performance would be very poorly in which `upsert_all` is preferred.

We are proposing on this PR a method to expire the cache when you don't have a direct access to an ActiveRecord instance. The method is attached to the `Attribute` class and we can access it like this:
```ruby
# item.rb
cache_attribute :name, by: :id
```
```ruby
# item_management.rb in the project
NameOfScope::Item.cache_indexes.each do |index|
    index.expire_by_key_value({ id: 1 })
end
```

We decided to pass a hash for the case of when we have an attribute cached by multiple keys:
```ruby
# item.rb
cache_attribute :name, by: [:id, :item_id]
```
```ruby
# item_management.rb in the project
NameOfScope::Item.cache_indexes.each do |index|
    index.expire_by_key_value({ id: 1, item_id:10 })
end
```

This also lets us have this case:
```ruby
# item.rb
cache_attribute :name, by: :id
cache_attribute :name, by: :other_property
cache_attribute :name, by: [:id, :item_id]
```
```ruby
# item_management.rb in the project
NameOfScope::Item.cache_indexes.each do |index|
    index.expire_by_key_value({ id: 1, item_id: 10, other_property: 11 })
end
```
We made this the most flexible possible to be able to handle these different possibilities as well as to avoid expiring the cache and missing one key, the attribute class will raise an exception letting you know that there is a missing field.

If the update changes the value of the key, then pluck should be used to get the current values, and the expiration call should include both the new and old values of the keys.

There is also some discussion about these changes, the main questions that comes out from these changes are:
* Should we also add methods to expire `expire_<attribute>_by_<key>`? In my opinion it could add a possible problem where you only expire one key of the object and not the other one, but we believe it is a good discussion to have.
* Should the gem have a method that does the `NameOfScope::Item.cache_indexes.each`

@Shopify/delivery-platform

### DIFF
--- a/lib/identity_cache/cached/attribute.rb
+++ b/lib/identity_cache/cached/attribute.rb
@@ -47,6 +47,19 @@ module IdentityCache
         end
       end
 
+      def expire_by_key_value(key_values)
+        missing_keys = missing_keys(key_values)
+        unless missing_keys.empty?
+          raise MissingKeyName,
+            "#{model.name} attribute #{alias_name} expire_by_key_value - "\
+            "required fields: #{key_fields.join(", ")}. "\
+            "missing: #{missing_keys.join(", ")}"
+        end
+
+        key = cache_key_by_values(key_values)
+        IdentityCache.cache.delete(key)
+      end
+
       def cache_key(index_key)
         values_hash = IdentityCache.memcache_hash(unhashed_values_cache_key_string(index_key))
         "#{model.rails_cache_key_namespace}#{cache_key_prefix}#{values_hash}"
@@ -100,6 +113,11 @@ module IdentityCache
         cache_key_from_key_values(new_key_values)
       end
 
+      def cache_key_by_values(key_values)
+        new_key_values = key_fields.map { |field| key_values[field] }
+        cache_key_from_key_values(new_key_values)
+      end
+
       def old_cache_key(record)
         old_key_values = key_fields.map do |field|
           field_string = field.to_s
@@ -117,6 +135,10 @@ module IdentityCache
 
       def fetch_method_suffix
         "#{alias_name}_by_#{key_fields.join("_and_")}"
+      end
+
+      def missing_keys(key_values)
+        key_fields - key_values.keys
       end
     end
   end
--- a/test/expire_by_key_values_test.rb
+++ b/test/expire_by_key_values_test.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+require "test_helper"
+
+class ExpireByKeyValuesTest < IdentityCache::TestCase
+  NAMESPACE = IdentityCache::CacheKeyGeneration::DEFAULT_NAMESPACE
+
+  def setup
+    super
+    AssociatedRecord.cache_attribute(:name)
+    AssociatedRecord.cache_attribute(:name, by: :item_id)
+    AssociatedRecord.cache_attribute(:name, by: [:id, :item_id])
+
+    parent = Item.create!(title: "bob")
+    parent.associated_records.create!(name: "foo")
+    IdentityCache.cache.clear
+  end
+
+  def test_expire_by_key_values
+    assert_queries(3) do
+      assert_equal("foo", AssociatedRecord.fetch_name_by_id(1))
+      assert_equal("foo", AssociatedRecord.fetch_name_by_item_id(1))
+      assert_equal("foo", AssociatedRecord.fetch_name_by_id_and_item_id(1, 1))
+    end
+
+    assert_queries(0) do
+      assert_equal("foo", AssociatedRecord.fetch_name_by_id(1))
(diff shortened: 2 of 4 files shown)


lib/identity_cache/cached/attribute.rb
```diff
@@ -47,6 +47,15 @@ def expire(record)
         end
       end
 
+      def expire_by_key_value(key_values)
+        key_fields.each do |key| 
+          return unless key_values.key?(key)
```
Larochelle, line 52: Should we raise instead?

lib/identity_cache/cached/attribute_by_multi.rb
```diff
@@ -31,6 +42,8 @@ def load_from_db_where_conditions(key_values)
         Hash[key_fields.zip(key_values)]
       end
 
+
```
Larochelle, line 45: 👾

lib/identity_cache/cached/attribute.rb
Hectorhammett: I also thought about that, but what happens if you have:
```ruby
cache_attribute :tracked, :id
cache_attribute :tracked, :sku
cache_attribute :tracked, :another_thing
```
And for whatever reason you want to expire just `id` and `sku`, then this would raise. Why would you want to do that? I don't really believe there is a good reason, but I am trying to make it as flexible as possible.

That being said, we could make this method being very specific to expire ALL of the indexes and the contract being you have to provide all the different indexes.

Larochelle: I was thinking that in most cases you want to expire all the caches, and perhaps we should optimize for ensuring we don't forget one.

Hectorhammett: We could definitely raise an error then..

dylanahsmith (changes requested):
> Using the `cache_attribute` method can be complicated on the case that you expire it when you utilize an `upsert_all` or `update_all` method on ActiveRecord as they do not return an instance of ActiveRecord.

It isn't complicated *because* you don't have an ActiveRecord instance.  The API is coupled to Active Record instances and Active Record callbacks because it provides the required data to both invalidate the old and new key if the cache index key columns change.

Even if everyone you tagged on the PR understands the problem, it is still important to make it clear what the scope of the problem that the PR actually addresses and it can be helpful context to anyone coming across the PR (e.g. through git archeology or issue/PR searching).

`upsert_all` and `update_all` are really just doing underlying database statements that don't load the rows.  That doesn't mean we couldn't load the rows before those statements to get the Active Record objects.  Was this considered as a way to get the old values?

Does your specific problem have a more limited scope, such as knowing that the cache indexes keys won't be modified?

> We decided to add methods for this use case using `expire_by_key_value` and adding dynamic methods for `expire_<attribute>_by_<key>` to be able to expire the cache.

This doesn't really explain the solution.  What are the thoughts behind that decision?  Why did you add more than one way to expire these caches?  How do you intend for these methods to be used to actually solve the problem?

The meta-programming defined methods to invalidate individual cache indexes seem like they could end up with code implicitly coupled to the set of cache indexes, such that a new one being added could result in it not being invalidated by existing code doing these manual cache invalidations.  I like how the approach taken with `expire_by_key_value` seems like it helps avoid this problem, although I still don't like how it hides the old key problem.

Feel free to answer questions in the form of documentation.  I'm also very concerned about how this could be mis-used, so documentation that helps guide the library user would be helpful.

test/expire_by_key_values_test.rb
```diff
@@ -0,0 +1,70 @@
...
+      assert_equal("foo", AssociatedRecord.fetch_name_by_id_and_item_id(1, 1))
+    end
+
+    assert_queries(0) do
+      assert_equal("foo", AssociatedRecord.fetch_name_by_id(1))
+      assert_equal("foo", AssociatedRecord.fetch_name_by_item_id(1))
+      assert_equal("foo", AssociatedRecord.fetch_name_by_id_and_item_id(1, 1))
+    end
+
+    key_values = {
```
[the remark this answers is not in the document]
dylanahsmith: The rubocop failure was a bug and the fix hasn't been released yet.  I opened https://github.com/Shopify/identity_cache/pull/496 to fix that.

PR #500, opened 2021-06-14
Add support for manual expiration for after direct DB manipulation

dylanahsmith:
cc @Larochelle & @Hectorhammett 

I've started an alternative to https://github.com/Shopify/identity_cache/pull/495 (and the existing `expire_primary_key_cache_index` class method) since I think we need a way to expire caches that doesn't make assumptions about what caches need to be expired.  Otherwise, it is easy to miss caches that the developer using this manual expiration didn't think about (e.g. from embedded associations) or ones added in the future.

This PR adds `cache_indexed_columns` class method that can be used to get the columns that are needed for cache expiration (see its documentation for details).  After that data is queried and put into the form of a hash, it can then be used with `expire_cache_for_delete` or `expire_cache_for_update` class methods.  I've also added `expire_cache_for_insert`, which could just be used with the data to insert.

Note that I haven't yet tested or fully documented this code.


lib/identity_cache/parent_model_expiration.rb
```diff
@@ -28,6 +29,22 @@ def install_pending_parent_expiry_hooks(model)
...
       end
 
+      def parent_expiration_entries
+        ParentModelExpiration.install_pending_parent_expiry_hooks(cached_model)
+        _parent_expiration_entries
+      end
+
+      def check_for_unsupported_parent_expiration_entries
+        return unless parent_expiration_entries.any?
+        msg = "Unsupported manual expiration of record embedded in parent associations:\n"
```
dylanahsmith, line 39: @Hectorhammett will this be needed for your use case?

Hectorhammett: This should not cause any problems with our use.

Hectorhammett (approved):

lib/identity_cache/query_api.rb
```diff
@@ -9,6 +9,56 @@ def prefetch_associations(includes, records)
...
+
+      def expire_cache_for_update(old_indexed_values, changes)
+        if primary_cache_index_enabled
+          id = old_indexed_values.fetch(primary_key.to_sym)
+          expire_primary_key_cache_index(id)
+        end
+        cache_indexes.each do |cached_attribute|
+          cached_attribute.expire_for_update(old_indexed_values, changes)
+        end
+        parent_expiration_entries.each do ||
```
Larochelle, line 43: Missing `end` for this block?

dylanahsmith: This line was meant to be deleted.  I'll reiterate that

> Note that I haven't yet tested or fully documented this code

I opened it early as a draft PR, since initially I was just trying to review and make recommendations for https://github.com/Shopify/identity_cache/pull/495 then decided code would better communicate this... but I didn't want to delay that communication in order to get this PR working.

I've pushed a fix to avoid breaking the existing tests... but it still needs tests for the new code paths.

Larochelle (approved):
🙏🏻

lib/identity_cache/cached/attribute.rb
```diff
@@ -34,19 +34,33 @@ def fetch(db_key)
...
         end
       end
 
+      def expire_for_values(values_hash)
+        key_values = key_fields.map { |name| values_hash.fetch(name) }
+        IdentityCache.cache.delete(cache_key_from_key_values(key_values))
+      end
+
+      def expire_for_update(old_values_hash, changes)
+        expire_for_values(old_values_hash, changes)
```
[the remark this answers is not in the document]
dylanahsmith: Oh yeah, this should be `expire_for_values(old_values_hash)`

test/cache_manual_expire_test.rb
```diff
@@ -0,0 +1,89 @@
...
+      error = assert_raises do
+        AssociatedRecord.expire_cache_for_update(old_values, new_values)
+      end
+
+      assert_equal(expected_error_message, error.message)
+    end
+
+    def test_expire_cache_for_insert
+      id = 1
+      AssociatedRecord.fetch_name_by_id(id)
```
dylanahsmith, line 65: Why is there no assertion for what value is returned by this?  Also, why is this making an assertion for a hard code id?

Shouldn't this test be making the insertion after this using something like `insert_all` that skips callbacks?  It looks like this is loading a record that already exists

test/cache_manual_expire_test.rb
```diff
@@ -0,0 +1,89 @@
...
+    def test_expire_cache_for_udate_raises_when_a_hash_is_missing_an_index_key
+      expected_error_message = "key not found: :id"
+      old_values = {
+        name: "foo",
+      }
+      new_values = {
+        name: "bar",
+      }
+
+      error = assert_raises do
```
dylanahsmith, line 56:
```suggestion
      error = assert_raises(KeyError) do
```

PR #502, opened 2021-09-08
Do not raise scope error if IdentityCache is disabled

kenzan100:
## Why

`should_use_cache?` method acts as a guard when the model/DB is in an undesirable state to use the IDC.

We're overriding this method to manipulate when to use model IDC conditionally.
I think raising an Rails scope error when cache is not being used, is misleading.

## Reviews

Is my understanding of `should_use_cache?` correct? Will there be a case we want to raise this error, even when `should_use_cache?` is false?

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -22,6 +22,8 @@ module IdentityCache
       private
 
       def raise_if_scoped
+        return unless should_use_cache?
+
         if current_scope
           IdentityCache.logger.error("#{name} has scope: #{current_scope.to_sql} (#{current_scope.values.keys})")
           raise UnsupportedScopeError, "IdentityCache doesn't support rails scopes (#{name})"


dylanahsmith:
> We're overriding this method to manipulate when to use model IDC conditionally.
> I think raising an Rails scope error when cache is not being used, is misleading.

The caller should not be calling Identity Cache in a way that only works if it is using the database.  Not only does that seem misleading, but that likely means that another code path likely has a bug when the code actually does get used for caching.  If Identity Cache is used without the expectation to actually use the cache, then just use the database directly.

---

Active Record scopes are global state that are mostly used to make it easy to build scopes using model class methods.  Using the scope for a larger block of code makes the code harder to understand, since there ends up being an implicit global side effect that affects all the code used in that block.

The reason why this check was added was to prevent this global state from corrupting the cache, since it could end up filling the cache with the result of the scoped query, changing the behaviour of code fetching from the cache without that scope.  This is why Identity Cache opted out of being used in this way.

This also has the nice added benefit that it makes it clear when reading code calling Identity Cache that it isn't in the context of one of these scopes, so removes a possible source of complexity to have to even consider.  In fact, it has even caught global state leaks of scopes that had their scope cleanup ensure block interrupted by a `Timeout` interrupt.

PR #503, opened 2021-10-26
Added new method to safely fetch read only records

joelpinheiro:
**What are you trying to accomplish with this PR?**

As `with_fetch_read_only_records` is not thread-safe, the following happens:

1. T1 calls `with_fetch_read_only_records(false)`, value is set to false.
2. T2 calls `with_fetch_read_only_records(false)`, value is set to false
3. T1 loads user.
4. T1 resets `fetch_read_only_records` to the old value (true).
5. T2 loads user (which is set to read only)
6. T3 resets `fetch_read_only_records` to old value (true, so no change).

By not having the `fetch_read_only_records` thread-safed we might have every now and then `ActiveRecord::ReadOnlyRecord` errors.

**How does it do it?**

A new method was created to fetch read only records flag safely:
```ruby
    def fetch_read_only_records
      v = Thread.current[:identity_cache_fetch_read_only_records]
      return v unless v.nil?
      fetch_read_only_records
    end
```

Which will be used in `with_fetch_read_only_records`.

I think we could simply use `IdentityCache.fetch_read_only_records = false` in the initializer but we want to set the IDC fetch strategy granularly.

### DIFF
--- a/lib/identity_cache.rb
+++ b/lib/identity_cache.rb
@@ -73,7 +73,7 @@ module IdentityCache
   # Fetched records are not read-only and this could sometimes prevent IDC from
   # reflecting what's truly in the database when fetch_read_only_records is false.
   # When set to true, it will only return read-only records when cache is used.
-  mattr_accessor :fetch_read_only_records
+  mattr_writer :fetch_read_only_records
   self.fetch_read_only_records = true
 
   class << self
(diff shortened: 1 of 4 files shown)


lib/identity_cache.rb
```diff
@@ -191,10 +191,16 @@ def fetch_multi(*keys)
...
       yield
     ensure
-      self.fetch_read_only_records = old_value
+      Thread.current[:identity_cache_fetch_read_only_records] = old_value
+    end
+
+    def fetch_read_only_records
+      v = Thread.current[:identity_cache_fetch_read_only_records]
+      return v unless v.nil?
+      @fetch_read_only_records
```
casperisfine, line 203: `mattr_writer` uses a class variable (`@@fetch_read_only_records`).

PR #507, opened 2021-11-29
Add spin support

dylanahsmith:
This repo already had `dev` support for Shopify employees.  It seems like it should also support using spin.

Note that this doesn't include postgres integration


garethson (approved):
Thanks!

PR #508, opened 2021-11-30
Fix `IdentityCache.fetch_read_only_records` default value

casperisfine:
Ref: https://github.com/Shopify/identity_cache/pull/503

`mattr_writer` uses a class variable internall, so `@fetch_read_only_records`
was refering to a never defined variable, so always falsy.

cc @dylanahsmith @joelpinheiro @tgwizard


tgwizard (approved):
oh, right. 

Perhaps there should be a test that `IdentityCache.fetch_read_only_records` is true by default, especially given that this has changed from a simple attribute to a more complex method?

casperisfine:
```
MemoizedAttributesTest#test_memoization_with_fetch_multi_should_not_break_dirty_tracking_with_filled_cache:
ActiveRecord::ReadOnlyRecord: Item is marked as readonly
```

There's something leaking in the test suite, but not sure what.

lib/identity_cache.rb
```diff
@@ -190,7 +190,7 @@ def fetch_multi(*keys)
     end
 
     def with_fetch_read_only_records(value = true)
-      old_value = fetch_read_only_records
+      old_value = Thread.current[:identity_cache_fetch_read_only_records]
```
casperisfine, line 193: The leak was here, after the first invocation `Thread.current[:identity_cache_fetch_read_only_records]` was reset to `true` instead of `nil`.

PR #519, opened 2022-08-12
Update rubocop-shopify to fix CI

dylanahsmith:
Rubocop was failing in CI with

```
Error: The `Gemspec/DateAssignment` cop has been removed. Please use `Gemspec/DeprecatedAttributeAssignment` instead.
(obsolete configuration found in /opt/hostedtoolcache/Ruby/3.0.4/x64/lib/ruby/gems/3.0.0/gems/rubocop-shopify-2.6.0/rubocop.yml, please update it)
```

So i updated rubocop-shopify, fixed some violations and added `NewCops: disable` to explicitly opt-out of new cops (so they aren't verbosely listed).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -3,6 +3,7 @@ inherit_gem:
 
 AllCops:
   TargetRubyVersion: 2.5
+  NewCops: disable
 
 Layout/BeginEndAlignment:
   EnforcedStyleAlignWith: start_of_line


dylanahsmith:
Ignore the CI / Rails edge failure, since #518 will address that.

gmcgibbon (approved):
Thanks!

PR #523, opened 2022-11-10
Return useful value from 'expire_cache'

driv3r:
It's possible for the underlying cache to fail a `write` operation which should set cache to deleted state across all the cache indexes. When this happens, cache indexes would be in inconsistent state, i.e. primary index could return `DELETED` value, as expected while other index might still have stale values. It could also happen that all writes fail and then we have stale cache when we want to refresh it.

So far we returned hardcoded `true` value from `expire_cache` method. After tracing `cache_attribute.expire(self)` method calls, I found out that the value returned is either:
- `truthy` if underlying connection doesn't throw an exception
- or `false` in case of the exception throwed by the underlying connection (Based on ActiveSupport Cache Stores API).

This PR modifies `expire_cache` to return `true` only when all the `expire` calls underneath return a truthy value and `false` when at least one of the calls fails. This should give means for knowing whenever the `expire_cache` call has to be repeated.

### DIFF
--- a/lib/identity_cache/query_api.rb
+++ b/lib/identity_cache/query_api.rb
@@ -171,10 +171,10 @@ module IdentityCache
       super
     end
 
-    # Invalidate the cache data associated with the record.
+    # Invalidate the cache data associated with the record. Returns `true` on success,
+    # `false` otherwise.
     def expire_cache
-      expire_attribute_indexes
-      true
+      expire_attribute_indexes.all?
     end
 
     # @api private
@@ -186,7 +186,7 @@ module IdentityCache
     private
 
     def expire_attribute_indexes # :nodoc:
-      cache_indexes.each do |cached_attribute|
+      cache_indexes.map do |cached_attribute|
         cached_attribute.expire(self)
       end
     end
--- a/test/expire_cache_test.rb
+++ b/test/expire_cache_test.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class ExpireCacheTest < IdentityCache::TestCase
+  def setup
+    super
+
+    Item.cache_index(:title, unique: true)
+    Item.cache_index(:id, :title, unique: true)
+
+    @record = Item.new
+    @record.id = 1
+    @record.title = "bob"
+
+    Spy.on(Item.cached_primary_index, :load_one_from_db).and_return do
+      @record
+    end
+  end
+
+  def test_expire_cache_hit
+    assert_equal(@record, Item.fetch(1))
+    refute_equal(IdentityCache::DELETED, backend.read(@record.primary_cache_index_key))
+    assert(@record.expire_cache)
+    assert_equal(IdentityCache::DELETED, backend.read(@record.primary_cache_index_key))
+  end
+
+  def test_expire_cache_on_failure
+    assert_equal(@record, Item.fetch(1))
+    refute_equal(IdentityCache::DELETED, backend.read(@record.primary_cache_index_key))
+
+    Spy.on(backend, :write_entry).and_return { false }
+
+    refute(@record.expire_cache)
+    refute_equal(IdentityCache::DELETED, backend.read(@record.primary_cache_index_key))
+  end
+end


test/expire_cache_test.rb
```diff
@@ -0,0 +1,37 @@
...
+  def test_expire_cache_on_failure
+    assert_equal(@record, Item.fetch(1))
+    refute_equal(IdentityCache::DELETED, backend.read(@record.primary_cache_index_key))
+
+    Spy.on(backend, :write_entry).and_return { false }
+
+    refute(@record.expire_cache)
+    refute_equal(IdentityCache::DELETED, backend.read(@record.primary_cache_index_key))
+  end
+end
```
dylanahsmith, line 37: This is missing a test for `expire_cache` returning `false` when `expire_primary_index` fails to invalidate a cache key.

This is also missing tests for when embedded cache associations are expired.  That is done by `expire_parent_caches`, but it looks like that isn't called as part of `expire_cache`.  I suppose that would have been fine before when `expire_cache` was just an internal API, but now that it is public, we should probably move the call to `expire_parent_caches` from `_run_commit_callbacks` into `expire_cache` for correctness.

test/expire_cache_test.rb
```diff
@@ -0,0 +1,37 @@
...
+    refute_equal(IdentityCache::DELETED, backend.read(@record.primary_cache_index_key))
+    assert(@record.expire_cache)
+    assert_equal(IdentityCache::DELETED, backend.read(@record.primary_cache_index_key))
+  end
+
+  def test_expire_cache_on_failure
+    assert_equal(@record, Item.fetch(1))
+    refute_equal(IdentityCache::DELETED, backend.read(@record.primary_cache_index_key))
+
+    Spy.on(backend, :write_entry).and_return { false }
```
dylanahsmith, line 32: `spy` isn't a widely used gem at Shopify, so I would prefer to limit usage on it to where integration testing or mocha aren't suitable.  In this case, it isn't even being used to spy on a method call without interfering with the call, it is just stubbing the call, which mocha could be used for.

Although, even better would be to avoid any fake to give confidence that this works as expected, which could be done by using a cache adapter configured with an open port.  I've done that in the past at Shopify with the following helper method

```ruby
module RandomOpenPort
  def self.find
    socket = Socket.new(:INET, :STREAM)
    socket.bind(Addrinfo.tcp("127.0.0.1", 0))
    socket.local_address.ip_port
  ensure
    socket.close if socket
  end
end
```

dylanahsmith (dismissed):

lib/identity_cache/query_api.rb
```diff
@@ -186,7 +186,7 @@ def was_new_record? # :nodoc:
     private
 
     def expire_attribute_indexes # :nodoc:
-      cache_indexes.each do |cached_attribute|
+      cache_indexes.map do |cached_attribute|
```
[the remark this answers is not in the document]
driv3r: just using `all?` caused early exit on the first falsely result, I've used `reduce` instead with sum being a second element in order to gather all results but without the need to allocate new array

test/expire_cache_test.rb
driv3r: I've moved expire parent caches to `expire_cache` method and include it in result as well

test/expire_cache_test.rb
dylanahsmith: Good point.  That seems like a good reason to use mocking.  I'd mostly just prefer using mocha over spy for stubbing.  A mocked server could work, but it might be a heavier weight solution than what is needed to test the short circuiting logic.

test/cache_invalidation_test.rb
```diff
@@ -211,4 +211,109 @@ def test_cache_not_expired_until_after_transaction
...
+    assert_equal(IdentityCache::DELETED, read_entity)
+  end
+
+  def test_expire_cache_with_primary_key_only_on_failure
+    assert_equal(@record, Item.fetch(1))
+    refute_equal(IdentityCache::DELETED, read_entity)
+
+    IdentityCache.cache_backend = CacheConnection.unconnected_cache_backend
+
+    refute(@record.expire_cache)
```
dylanahsmith, line 228: I think we want to use the normal cache connection after the expiring the cache for `read_entity` on the following line.  Otherwise, `refute_equal(IdentityCache::DELETED, read_entity)` could just pass because `read_entity` is returning `nil` when the cache isn't present.

There are a few places where `IdentityCache.cache_backend = CacheConnection.unconnected_cache_backend` is done that make the same mistake, so that may as well be done in a block method that resets the connection.

```ruby
  def with_unconnected_cache_backend
    IdentityCache.cache_backend = CacheConnection.unconnected_cache_backend
    yield
    IdentityCache.cache_backend = backend
  end
```

```suggestion
    with_unconnected_cache_backend do
      refute(@record.expire_cache)
    end
```

dylanahsmith (commented):
One concern with the tests, otherwise, looks good to me

test/cache_invalidation_test.rb
driv3r: @dylanahsmith to be honest in this particular examples it doesn't make much sense the `unconnected_backend`. It's fine to use it once to see if true/false is returned, but then it makes more sense to stub the server to mimic a single fluke, if we switch it back and forth it doesn't make sense, as we now that the `expire_cache` hit a completely different backend anyway, so there's no reason to believe that the initial values would be different. 

I'll look a little bit more into this, and fallback to stubs if I don't find anything suitable.

dylanahsmith: I'm not saying this is a replacement for stubbing.  Just that where we do use the unconnected_cache_backend, the assertions that follow should make sense.  For the code I was commenting on, it wasn't clear what those `refute_equal` statements were testing.

dylanahsmith:
> That said, from a debug perspective, a stupid question: because we are using a lot of `&&` to see if the delete was successful, we lose the information about where the failure came from. I suppose this is OK for the time being?

For debugging, we would probably want to look at the logs.  The failed delete should be logged in [MemoizedCacheProxy#delete](https://github.com/Shopify/identity_cache/blob/c77fdf0f04fe2895ce3099af2e7de233e5a1c212/lib/identity_cache/memoized_cache_proxy.rb#L67).

The boolean response is meant to be for being able to programmatically handle the failed deletion, where it is safer to treat the cache expiry as failed if invalidating any individual keys fail.  I would expect that the error handling behaviour that would response to this sort of failure would just retry the call to `expire_cache` at a later point.

lib/identity_cache/parent_model_expiration.rb
```diff
@@ -45,8 +45,9 @@ def lazy_hooks
     def expire_parent_caches
       parents_to_expire = Set.new
       add_parents_to_cache_expiry_set(parents_to_expire)
-      parents_to_expire.each do |parent|
-        parent.expire_primary_index if parent.class.primary_cache_index_enabled
+      parents_to_expire.select! { |parent| parent.class.primary_cache_index_enabled }
+      parents_to_expire.reduce(true) do |all_expired, parent|
+        parent.expire_primary_index && all_expired
```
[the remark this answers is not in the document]
driv3r: In our case not really, in the general one I think it would make sense to instrument it with something like AS Notifications, but it's outside of the current scope

test/helpers/mocked_cache_backend.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def stub_call(key, value)
+    stubbed_calls << [key, value]
+  end
+
+  def stubbed_calls
+    @stubbed_calls ||= []
+  end
+end
```
driv3r, line 22: @dylanahsmith usually with rspec I would go with something like 
```ruby
expect(backend).to receive(:write).and_return(false).once
expect(backend).to receive(:write).with(...).and_call_original
```

but from what I've checked mocha doesn't have anything like this, so I've decided to go with simple delegator instead.

test/cache_invalidation_test.rb
```diff
@@ -211,4 +211,166 @@ def test_cache_not_expired_until_after_transaction
...
+    refute_equal(IdentityCache::DELETED, read_entity)
+    assert(@record.expire_cache)
+    assert_equal(IdentityCache::DELETED, read_entity)
+  end
+
+  def test_expire_cache_with_primary_key_only_on_failure
+    assert_equal(@record, Item.fetch(1))
+    refute_equal(IdentityCache::DELETED, read_entity)
+
+    with_cache_backend do
```
dylanahsmith, line 226: This does the opposite of what it looks like from the callsite. It would be more readable to explicitly use the CacheConnection.unconnected_cache_backend argument

test/cache_invalidation_test.rb
driv3r: This makes sense! I've also adjusted rest of the tests to assert for correct value instead of refuting `DELETED` key. Sorry @dylanahsmith for a lot of back and forth, I'm still new to minitest approach 🙇‍♂️

dylanahsmith (approved):

PR #531, opened 2023-03-22
Leave embedded associations preloaded when should_use_cache is false

cbothner:
Our background jobs run against a read-replica database, and so IDC is disabled in order to avoid inconsistency stemming from the cache being "ahead" of the database. But today this results in N+1 queries when resources when `fetch` is used on resources with embedded associations.

This is how that comes to pass:
1. Set `should_use_cache` to false
2. Fetch a resource with embedded associations
3. IDC is forced to load from the database. It uses the preloads that it normally uses on a cache miss. 
    https://github.com/Shopify/identity_cache/blob/328b52fa9746105b71f079caa8fdbc3dd30804ba/lib/identity_cache/cached/primary_index.rb#L87-L89
4.  IDC sets up its embedded associations with the same behavior as if `should_use_cache` were true
    1. It populates the cached association
    2. and resets the bare AR association, per the reasoning in https://github.com/Shopify/identity_cache/pull/378
         https://github.com/Shopify/identity_cache/blob/328b52fa9746105b71f079caa8fdbc3dd30804ba/lib/identity_cache/query_api.rb#L88
    3. It then reads the cached association in order to set up the children's embedded associations
         https://github.com/Shopify/identity_cache/blob/328b52fa9746105b71f079caa8fdbc3dd30804ba/lib/identity_cache/query_api.rb#L101-L102
    4. But because `should_use_cache` is false, this bypasses the cached association and instead hits the DB N times
         https://github.com/Shopify/identity_cache/blob/328b52fa9746105b71f079caa8fdbc3dd30804ba/lib/identity_cache/cached/recursive/association.rb#L28

Even if IDC did not trigger the N queries internally, application code that uses the cached associations of a model will trigger the queries. Application code that uses embedded associations should be able to assume that they are eager loaded.

This PR proposes that when `should_use_cache` is false, the resources should not have their embedded associations hydrated using the normal logic. Instead, the instances should behave as if they were loaded with `find` and not `fetch`, with the `includes` that are necessary to meet application code's expectations for embedded associations.

cc @hovo @MatWrz

### DIFF
--- a/test/fetch_test.rb
+++ b/test/fetch_test.rb
@@ -342,47 +342,6 @@ class FetchTest < IdentityCache::TestCase
     end
   end
 
-  def test_with_should_use_cache_true_fetched_record_does_not_have_embedded_associations_preloaded_on_cache_miss
-    Item.cache_has_many(:associated_records, embed: true)
-    Item.cache_has_many(:normalized_associated_records)
-
-    @record.associated_records.build
-    @record.associated_records.build
-    @record.normalized_associated_records.build
-    @record.save
-
-    assert_memcache_operations(1) do
-      assert_queries(3) do
-        item = Item.fetch_by_id(@record.id)
-
-        refute(item.association(:associated_records).loaded?)
-        refute(item.association(:normalized_associated_records).loaded?)
-      end
-    end
-  end
-
-  def test_with_should_use_cache_true_fetched_record_does_not_have_embedded_associations_preloaded_on_cache_hit
-    Item.cache_has_many(:associated_records, embed: true)
-    Item.cache_has_many(:normalized_associated_records)
-
-    @record.associated_records.build
-    @record.associated_records.build
-    @record.normalized_associated_records.build
-    @record.save
-
-    # warm the cache
-    Item.fetch_by_id(@record.id)
-
-    assert_memcache_operations(1) do
-      assert_queries(0) do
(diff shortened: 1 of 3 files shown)


test/fetch_test.rb
```diff
@@ -342,6 +342,68 @@ def test_respects_should_use_cache_on_record
...
+
+    Item.stubs(:should_use_cache?).returns(false)
+
+    assert_memcache_operations(0) do
+      assert_queries(2) do
+        item = Item.fetch_by_id(@record.id)
+
+        assert(item.association(:associated_records).loaded?)
+        refute(item.association(:normalized_associated_records).loaded?)
+      end
```
dylanahsmith, line 403: Avoid making these blocks larger than they need to be, so the test is clearer (e.g. don't need to read past that to see if the assertion lines are relevant to the query count)

```suggestion
      item = assert_queries(2) do
        Item.fetch_by_id(@record.id)
      end

      assert(item.association(:associated_records).loaded?)
      refute(item.association(:normalized_associated_records).loaded?)
```

hovo (commented):
I assume we also need to create a new release after these changes are merged?

dylanahsmith:
> I assume we also need to create a new release after these changes are merged?

We could test the change against the git commit, but we would want to release the change to fix it for any other apps that may be affected by this.

PR #536, opened 2023-04-12
fix: `fetch_multi_by` bug for WHERE IN optimization

cribbles:
Fixes a bug introduced in https://github.com/Shopify/identity_cache/pull/534 for `#fetch_multi_by` invocations having a single field with distinct values.

The diff itself should make the bug pretty straightforward. This PR also adds a regression test against the case, which ensures we're now testing against each clause in the `case` statement.

I believe bumping to `1.4.1` should be germane in this case because the bug was associated with new behavior introduced in a major update, i.e., no client of this library should have been affected by the changes in #534 via a version bump alone.

Please note that CI will surely fail on the Rails edge step for unrelated reasons in the fetch test suite. This is a problem on `main`, too; you can verify it yourself by installing the Rails edge bundle and running `dev test test/fetch_test.rb`.

### DIFF
--- a/lib/identity_cache/cached/attribute_by_multi.rb
+++ b/lib/identity_cache/cached/attribute_by_multi.rb
@@ -84,7 +84,7 @@ module IdentityCache
           # This results in a single "WHERE field IN (values)" statement being
           # produced from a single query.
           field_idx = other_field_indexes.first
-          field_name = key_fields[i]
+          field_name = key_fields[field_idx]
           field_values = keys.map { |key| key[field_idx] }
           (common_query || unsorted_model).where(field_name => field_values)
         else
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,35 +5,42 @@
 ## 1.4.0
 
 ### Features
+
 - Add `fetch_multi_by` support for composite-key indexes. (#534)
 
 ## 1.3.1
 
 ### Fixes
+
 - Remove N+1 queries from embedded associations when using `fetch` while `should_use_cache` is false. (#531)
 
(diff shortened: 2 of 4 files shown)


lib/identity_cache/cached/attribute_by_multi.rb
```diff
@@ -84,7 +84,7 @@ def load_multi_rows_query(keys)
           # This results in a single "WHERE field IN (values)" statement being
           # produced from a single query.
           field_idx = other_field_indexes.first
-          field_name = key_fields[i]
```
dylanahsmith, line 87: Oh, I didn't realize this code path wasn't tested.  This does indeed lead to a undefined local variable error with the regression test and without this change.

dylanahsmith:
Rails edge CI failures are from https://github.com/rails/rails/pull/47864 and aren't related to this PR.  The tests should probably be fixed to avoid assuming that the unsaved record objects can be compared against the ones returned from the cache fetch with a mocked response.

lib/identity_cache/cached/attribute_by_multi.rb
cribbles: Yeah, this is something I thought we caught in [this test](https://github.com/Shopify/identity_cache/blob/4bcb03096dfe5bf9f083e3bcff494039016dc140/test/fetch_multi_by_test.rb#L119-L126) but it's not structured in such a way as to hit the path -- would be nice to find a non-hacky way of running a `query.to_sql`-like check to _ensure_ all these paths are hit in case parts of this suite are restructured in the future

PR #537, opened 2023-04-12
Fix tests on rails edge

dylanahsmith:
Follow-up to https://github.com/Shopify/identity_cache/pull/536#issuecomment-1505288908

> Rails edge CI failures are from [rails/rails#47864](https://github.com/rails/rails/pull/47864) and aren't related to this PR. The tests should probably be fixed to avoid assuming that the unsaved record objects can be compared against the ones returned from the cache fetch with a mocked response.

Instead of just comparing against the unsaved record instances, I just compared the object class and the id more explicitly.


cribbles (approved):
Thanks for investigating and taking care of this so quickly!

PR #538, opened 2023-04-12
Add assert_queries_sql test helper for asserting SQL of DB queries

dylanahsmith:
Follow-up to https://github.com/Shopify/identity_cache/pull/536#discussion_r1164156150

> Yeah, this is something I thought we caught in [this test](https://github.com/Shopify/identity_cache/blob/4bcb03096dfe5bf9f083e3bcff494039016dc140/test/fetch_multi_by_test.rb#L119-L126) but it's not structured in such a way as to hit the path -- would be nice to find a non-hacky way of running a `query.to_sql`-like check to _ensure_ all these paths are hit in case parts of this suite are restructured in the future

Specifically, we don't want to assert the SQL string against a hard coded string, since that would differ across databases, however, we could build a similar SQL string using the Active Record query build and `to_sql`.

There are also already test helpers for observing queries, which is used to assert the number of queries performed, so I added a new test helper to assert against the actual query SQL.  I've then used it in a sample test to ensure it works and to show how it can be used.

### DIFF
--- a/test/fetch_multi_by_test.rb
+++ b/test/fetch_multi_by_test.rb
@@ -186,14 +186,11 @@ class FetchMultiByTest < IdentityCache::TestCase
 
   private
 
-  def assert_query_from_fetch_multi_by(**options)
-    given_query = options[:given]
-    expected_query = options[:expect_query]
-    expected_entities = options[:returning]
+  def assert_query_from_fetch_multi_by(given:, expect_query:, returning:)
     result = assert_queries_sql(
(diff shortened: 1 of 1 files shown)


cribbles (approved):
Great approach to this!

I've applied it to the rest of the `fetch_multi` composite key tests in [this branch](https://github.com/Shopify/identity_cache/compare/add-assert-queries-sql-test-helper...cribbles/add-sql-test-coverage?expand=1), adding an additional helper method to cut down on some of the verbosity; feel free to cherry-pick one or both of the commits to taste.

cribbles:
LGTM 👍

Issue #29, opened 2013-04-17
has_and_belongs_to_many not working


standingwave:
I can't get a basic has_and_belongs_to_many association to work.  The setup is pretty simple:

``` ruby
class Site < ActiveRecord::Base
  include IdentityCache
  has_and_belongs_to_many :articles
  cache_has_many :articles, :inverse_name => :site
```

The inverse_name is there because I get a IdentityCache::InverseAssociationError but it doesn't make any difference.  Any ideas?

Gabriel

PR #548, opened 2023-11-29
Use foreign_key to reference parent association

slucaskim:


cbothner (commented):
I was interested in how specifically `has_and_belongs_to_many` associations made use of these two fields. Of course, this is irrelevant because this gem [doesn't support `has_and_belongs_to_many` associations](https://github.com/Shopify/identity_cache/issues/29). Just interesting

```ruby
class Store < ActiveRecord::Base
  has_and_belongs_to_many :customers
end

class Customer < ActiveRecord::Base
  has_and_belongs_to_many :stores
end

Store.reflect_on_association(:customers).foreign_key # => "store_id"
Store.reflect_on_association(:customers).association_foreign_key # => "customer_id"

cbothner (approved):

PR #550, opened 2023-12-11
Add should_fill_cache to a few places

drinkbeer:


test/readonly_test.rb
```diff
@@ -77,8 +78,10 @@ def assert_readonly_fetch
 
   def assert_readonly_fetch_multi
     cas_multi = Spy.on(backend, :cas_multi).and_call_through
+    write = Spy.on(backend, :write).and_call_through
     yield
     assert(cas_multi.has_been_called?)
+    refute(write.has_been_called?)
```
drinkbeer, line 84: We set the cache to read-only (aka, should_fill_cache is false), so we don't call `write` method.

drinkbeer:
> We should add tests for the two new code paths we introduced (if should_use_cache? is true or false in both cases).

We already have cases cover should_use_cache? to be true. I added a test to cover the case that should_use_cache? to be false.

camilo (approved):

PR #553, opened 2024-01-18
Do not call should_use_cache? on prefetched/embedded assocs

kirs:
`IdentityCache.should_use_cache?` can be rather expensive as it checks for open transactions. Consider this example:

```ruby
# ProductVariant belongs to Product
variants = ProductVariant.fetch_multi(1000_of_ids, includes: [:product])
variants.each |v|
  product = v.fetch_product
  # do something with the product
end
```

The following would essentially call `IdentityCache.should_use_cache?` at least 1001 times. That's 1001 times of checking for transaction to be open.

For real, we only need to check `should_use_cache?` in the top level `fetch`/`fetch_multi`. For all of those variants the product has already been prefetched, and there's little point calling `should_use_cache?` on each of those `fetch_product`.

This is accomplished by marking the record as "loaded by IDC". That's a marker signaling that `should_use_cache?` has already been checked.

`ProductVariant.find(1).fetch_product` would not have that market set, and it would still run `should_use_cache?` as expected.

### DIFF
--- a/lib/identity_cache/cached/primary_index.rb
+++ b/lib/identity_cache/cached/primary_index.rb
@@ -54,7 +54,10 @@ module IdentityCache
 
       def load_one_from_db(id)
         record = build_query(id).take
-        model.send(:setup_embedded_associations_on_miss, [record]) if record
+        if record
+          model.send(:setup_embedded_associations_on_miss, [record])
+          record.send(:mark_as_loaded_by_idc)
+        end
         record
       end
 
--- a/lib/identity_cache/cached/primary_index.rb
+++ b/lib/identity_cache/cached/primary_index.rb
@@ -66,6 +66,7 @@ module IdentityCache
 
         records = build_query(ids).to_a
         model.send(:setup_embedded_associations_on_miss, records)
+        records.each { |record| record.send(:mark_as_loaded_by_idc) }
         records.index_by(&:id)
       end
 
--- a/test/denormalized_has_many_test.rb
+++ b/test/denormalized_has_many_test.rb
@@ -171,17 +171,6 @@ class DenormalizedHasManyTest < IdentityCache::TestCase
     end
   end
 
-  def test_respect_should_use_cache_from_embedded_records
-    Item.fetch(@record.id)
-    AssociatedRecord.stubs(:should_use_cache?).returns(false)
-
-    assert_memcache_operations(1) do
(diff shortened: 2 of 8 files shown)


test/denormalized_has_many_test.rb
```diff
@@ -171,17 +171,6 @@ def test_returned_records_with_open_transactions_should_not_be_readonly
     end
   end
 
-  def test_respect_should_use_cache_from_embedded_records
```
kirs, line 174: This test was added waaay in the back: https://github.com/Shopify/identity_cache/commit/74d91534b61bf254bbf5969760afe1a1fbdb3965

It asserts on behaviour that's opposite from what this PR does, so removing it.

lib/identity_cache/cached/primary_index.rb
```diff
@@ -54,7 +54,10 @@ def cache_key(id)
 
       def load_one_from_db(id)
         record = build_query(id).take
-        model.send(:setup_embedded_associations_on_miss, [record]) if record
+        if record
+          model.send(:setup_embedded_associations_on_miss, [record])
+          record.send(:mark_as_loaded_by_idc)
+        end
```
drinkbeer, line 60: Should we do the same thing for `load_multi_from_db`?

drinkbeer: Here you call the `mark_as_loaded_by_idc` in the `load_one_from_db` method is based on the assumption that whenever the method is called, the IDC is filled, right?

drinkbeer:
Should the has_one also check the `loaded_by_idc?` mark first before calling the `should_use_cache?`?

https://github.com/Shopify/identity_cache/blob/main/lib/identity_cache/cached/reference/has_one.rb#L26

drinkbeer (approved):
LGTM. Ship after addressing some comments.

drinkbeer:
> The first is covered by the fact that class HasOne < Association and this PR modifies Association. I have a test to validate that.

> Reference::HasOne is basically for embed: :id. Since embedded ids (either with has_one or has_many) would involve another round trip to memcache, I've chose to keep those paths as they are. Basically, the scope of my PR is only embedded and prefetched associations. I've added tests to assert on those behaviour too. For those, there will still be should_use_cache? called on each model.

Thank you. I think it makes sense.

PR #556, opened 2024-02-09
Forward expiration options

TheDro:
Forward expiration_options through the CacheFetcher class so that public methods can use it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Exposed `expiration_options` on `CacheFetcher`
+
 ## 1.5.3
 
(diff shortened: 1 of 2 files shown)


TheDro:
I have signed the CLA!

danmayer:
Thanks for submitting the PR, we are trying to keep the interface and usage of Identity Cache simple and avoiding having additional complexity leak into the API.

The philosophy of the library is that all IDC writes are the same.

We are going to close this PR, if this kind of optimization is important to you all feel free to fork or patch.

danmayer:
yeah, that use case is what we were guessing you were looking to do. We don't have a need for that within our usage and want to restrict any API complexity as it is already a challenging library to debug.

So for your use case, applying minimal patches locally or maintaining a small fork would be the best path forward.

PR #557, opened 2024-02-11
Fix class name in AssociationError message

eugeneius:
`self` is the model here, so `self.class` is always "Class".


nvasilevski (approved):
Thank you!

PR #561, opened 2024-02-23
Adds support for models using compound primary keys

richardpenner:


lib/identity_cache/cached/attribute.rb
```diff
@@ -138,10 +149,15 @@ def field_types
 
       def cache_key_prefix
         @cache_key_prefix ||= begin
+          attribute_key_prefix = if attribute.is_a?(Array)
+            attribute.map(&:to_s).join("/")
```
gmac, line 153: Nit: you shouldn't need the intermediary string mapping. Symbols will coerce while being joined.

PR #562, opened 2024-02-27
Early return on empty IDs to fetch_multi

kirs:
In Shopify app, it happens often that a record has `cache_has_many :some_relation`, but that relation is empty.
On calls to `fetch_some_relation`, it happens that `fetch_multi` is called with an empty array.

Whenever that `fetch_multi([])` gets called, we go through a series of expensive checks like `IdentityCache.should_use_cache?`  -- while all of that is quite unnecessary and we can shirt-circuit that.

To illustrate it: a live profile that calls `fetch_media_references` 1000 times for an empty relation (which ends up being `fetch_multi([])`)

Screenshot 2024-02-27 at 12 33 41 PM


danmayer (approved):
nice find and fix

PR #563, opened 2024-03-15
Centralize Ruby Version to `.ruby-version`

george-ma:
## What are you trying to accomplish?
The `.ruby-version` file is the ecosystem standard for defining a Ruby version.  This PR adds the `.ruby-version` file, ensures a `required_ruby_version` is set, and removes all other references to Ruby in this repository, aligning it with the standard.

Verify that the changes in the PR meets the following requirements or adjust manually to make it compliant:
  - [x] `.ruby-version` file is present with the correct Ruby version defined
  - [x] A `required_ruby_version` in your gemspec is set
  - [x] There is no Ruby version present in the  `dev.yml` Ruby task (before: `- ruby: x.x.x`, after: `- ruby`)
  - [x] There is no Ruby version/requirement referenced in the `Gemfile` (no lines with `ruby  <some-version>`)
  - [x] A `Gemfile.lock` is built with the defined Ruby version
  - [x] The version of Rubocop installed is 1.61.0 or greater
  - [x] There is no `TargetRubyVersion` defined  in `rubocop.yml` 
  - [x] There is no Ruby argument  present in  `ruby/setup-ruby` Github Actions that do **not**  run on a Ruby matrix (no lines with `ruby: “x.x”`)

This PR will be merged if there isn't any activity after 4 weeks.


kirs (commented):
LGTM except `Gemfile.lock`. I'm not sure we want to track that one for a gem. What was your reason to add it?

PR #565, opened 2024-03-15
Add x86_64-linux as a supported platform

george-ma:
Follow up to: https://github.com/Shopify/identity_cache/pull/563

Need to add x86_64-linux to the Gemfile.lock (noted for future repos) - necessary for shipit


george-ma:
```
PLATFORMS
  ruby
```
would work too (`ruby`: Indicates that the gems are compatible with a standard Ruby platform) but from looking at other libraries that commit their Gemfile.lock, including x86_64-linux is pretty common
- https://github.com/Shopify/packwerk/blob/main/Gemfile.lock
- https://github.com/Shopify/rbi/blob/main/Gemfile.lock
- https://github.com/Shopify/oktakit/blob/main/Gemfile.lock

bettymakes:
@george-ma The details you expanded on in the comment are useful in the PR description itself. How about restructuring your description using some headings to help better communicate the problem you're solving and the approach you're taking? You can elaborate on alternative approaches in the description too.

Suggested headings as helpful prompts:
- What problem are you solving?
- What approach did you choose and why?

lastgabs (approved):

PR #566, opened 2024-03-18
Change order in if clause for should_use_cache?

kirs:
When records are preloaded from database using `Post.includes(:author).find(id)`, `fetch_author` will fallback to returning the object that was preloaded from the database.

However, it still goes through `association_klass.should_use_cache?` which started appearing on our profiles.

`association_klass.should_use_cache?` is the more expensive check so it's gotta be in the end of `if` clause.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+## 1.5.6
+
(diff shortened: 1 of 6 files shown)


bitwise-aiden (approved):

danmayer (approved):
good call, thanks for fixing this

PR #573, opened 2024-08-19
Use ActiveRecord::ConnectionAdapters::ConnectionPool#lease_connection with Rails 7.2

eiskrenkov:
Fix deprecation warning for Rails 7.2 coming from [here](https://github.com/rails/rails/blob/b60cf01e9d076d235fdb02db1072ea2a31c94dbe/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L295)

```
DEPRECATION WARNING: ActiveRecord::ConnectionAdapters::ConnectionPool#connection is deprecated
and will be removed in Rails 8.0. Use #lease_connection instead.
```

### DIFF
--- a/lib/identity_cache.rb
+++ b/lib/identity_cache.rb
@@ -137,7 +137,7 @@ module IdentityCache
           # When there are no open transactions, `current_transaction` returns a
           # special `NullTransaction` object that is unjoinable, meaning we will
           # use the cache.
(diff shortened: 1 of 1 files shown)


casperisfine:
Fixed in https://github.com/Shopify/identity_cache/pull/575

PR #624, opened 2026-02-27
Improve MemCacheStoreCAS patch to properly pass serialized entries to Dalli

byroot:
Instead it was relying on the fact that dalli will Marshal those, but it's much preferable to rely on the Active Support layer instead, given Dalli 4 deprecate Marshal and Dalli 5 no longer does it by default.

### DIFF
--- a/lib/identity_cache/mem_cache_store_cas.rb
+++ b/lib/identity_cache/mem_cache_store_cas.rb
@@ -4,6 +4,8 @@ require "dalli/cas/client" unless Dalli::VERSION > "3"
 
 module IdentityCache
   module MemCacheStoreCAS
+    Entry = ActiveSupport::Cache::Entry
+
     def cas(name, options = nil)
       options = merged_options(options)
       key = normalize_key(name, options)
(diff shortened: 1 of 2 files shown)


byroot:
cc @adrianna-chang-shopify @Edouard-chin (sorry not sure who watches the repo, and you are the last few to have merged / released :) )

lib/identity_cache/mem_cache_store_cas.rb
```diff
@@ -43,21 +45,12 @@ def cas_multi(*names, **options)
...
-            payload = options[:raw] ? entry.value.to_s : entry
+            entry = Entry.new(value, **options)
+            payload = serialize_entry(entry, **options)
             @data.with { |c| c.replace_cas(key, payload, cas_id, options[:expires_in].to_i, options) }
           end
         end
       end
     end
-
-    if ActiveSupport::Cache::MemCacheStore.instance_method(:deserialize_entry).arity == 1
```
byroot, line 54: That is no longer necessary as we require Rails >= 7.0

lib/identity_cache/mem_cache_store_cas.rb
```diff
@@ -14,8 +16,8 @@ def cas(name, options = nil)
             connection.cas(key, options[:expires_in].to_i, options) do |raw_value|
               entry = deserialize_entry(raw_value, raw: options[:raw])
```
Edouard-chin: Should we remove the `raw` option here too ?

byroot: Hum, yeah we should pass `**options`.

Edouard-chin (approved):

